PR #205, opened 2026-07-16
feat(client): drop unknown WebSocket frames without closing the connection

kartojal:
### Summary

WebSocket streams never close the connection or fail the session when the server sends a frame the SDK cannot parse. Unknown or invalid frames — new frame types, malformed known frames, or known frames carrying enum values the SDK does not model yet — are silently dropped, and the stream keeps delivering subsequent events.

No observer hook is exposed and the SDK does not log dropped frames: surfacing raw unmodeled frames would invite consumers to couple to untyped shapes the SDK has not had a chance to model yet.

### Behavior changes

| Scenario | Before | After |
|---|---|---|
| New frame type on any stream | Silently dropped | Unchanged (silently dropped, connection stays open) |
| Malformed known frame (CLOB/RTDS/sports/perps) | Silently dropped | Unchanged |
| Malformed known frame (RFQ) | Session failed with `TransportError`, all pending acks rejected | Dropped; session stays open, the waiting call fails through its ack timeout |
| Unknown frame type / non-object frame (RFQ) | Session failed with `TransportError` | Dropped; session stays open |
| Known frame with a status value the SDK does not model (RFQ execution update, CLOB trade/order, perps TP/SL) | RFQ: session failed; others: frame dropped | Dropped everywhere; session and connection stay open |
| Well-formed but uncorrelated RFQ error | Session failed | Unchanged (deliberate) |

### Changes

**`packages/client`**

- The RFQ quoter no longer fails the session on unparseable frames. It previously raised `TransportError('Invalid RFQ quoter message.')` and tore down the session, rejecting all pending acknowledgements; such frames are now dropped using the same single full-parse classification as every other stream (the separate known-type pre-check was removed). Pending calls whose acks never arrive fail through their existing ack timeout — the same `TimeoutError` consumers already handle for a silent server.
- Every stream manager drops frames that fail schema validation with a `continue`/`return`, keeping the connection and all subscriptions open. Control frames (heartbeat PONGs, perps request-echo acks) are filtered before classification as before.
- Status enums stay strict (`TradeStatus`, `UserOrderStatus`, `UserOrderEventType`, `RfqExecutionStatus`, perps TP/SL `st`): a known frame carrying an unrecognized status fails validation and is dropped like any other invalid frame. The pre-existing `RfqErrorCode` string fallback is unchanged.

**`packages/bindings`**

- Removed `RfqKnownInboundMessageSchema`. Its only consumer was the quoter's removed known-type pre-check, and the loose `{ type }` base was extended with its `type` field overwritten by every concrete inbound message schema — each schema now declares its own object shape directly.

### Tests

- Shared helper `expectDropsUnknownFrame` (`websockets/testing.ts`): sends an unrecognized frame followed by a valid one and asserts the valid frame's event is the next thing delivered — proving the unknown frame emitted nothing and the connection and subscription survived. Used across the CLOB market/user, RTDS, sports, and perps stream tests.
- Integration test: a malformed known RFQ frame (`ACK_RFQ_QUOTE` missing `quote_id`) is dropped while the valid ack sent afterwards still correlates and the quote resolves — the session survives.
- Regression tests: RFQ `execution_update`, CLOB trade, and perps TP/SL frames carrying an unrecognized status are dropped without closing the session; the next valid frame is still delivered.
- Perps tests assert request-id acknowledgements are not treated as unknown frames.

---

> [!NOTE]
> **Medium Risk**
> RFQ error handling changes from immediate session failure to silent drops on bad frames, which can delay failure until ack timeout; correlated RFQ errors and connection loss behavior are unchanged.
> 
> **Overview**
> WebSocket clients now **treat unparseable inbound frames as no-ops** instead of tearing down the socket or RFQ session. That covers new message types, malformed payloads, and known message shapes with enum values the SDK does not model yet; later frames still flow on the same connection.
> 
> The **RFQ quoter** is the main behavior change: it no longer raises `TransportError` and fails the session when `RfqQuoterInboundMessageSchema` rejects a frame. Those messages are dropped like on other streams; pending quote/cancel/confirm calls that never get a matching ack still fail via the existing **ack timeout** (`TimeoutError`). **Well-formed, correlated `RFQ_ERROR` responses still fail or reject the pending operation** as before; uncorrelated errors still fail the session.
> 
> In **`@polymarket/bindings`**, `RfqKnownInboundMessageSchema` is removed; each RFQ inbound schema is a standalone `z.object` (same validation, no shared base).
> 
> Tests add shared **`expectDropsUnknownFrame`** across CLOB market/user, RTDS, sports, and perps streams, plus RFQ integration cases for malformed acks, future execution statuses, and related regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2cc0180df834fed31b72dd94ce9b2addbdf1a0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/websockets/rfq/quoter.ts
```diff
@@ -274,16 +273,9 @@ class RfqWebSocketSession implements RfqSession, RfqEventController {
 
   #handleMessage(rawMessage: unknown): void {
     if (this.#closing !== undefined) return;
-    if (!RfqKnownInboundMessageSchema.safeParse(rawMessage).success) return;
```
cesarenaldi, line 277: We can probably remove the base schema RfqKnownInboundMessageSchema, it's a very loosely defined schema that is then extended and overwritten, so it does not even need to exist as such.

cesarenaldi (approved):
Just a nit comment, pre-approved.

packages/client/src/websockets/rfq/quoter.ts
kartojal: removed in commit [7633fad](https://github.com/Polymarket/ts-sdk/pull/205/commits/7633fadb31e5cd58242e58d1008b462047cdfd4b)

PR #198, opened 2026-07-14
fix(rfq): keep sessions open for new error codes

shazam8253:
## Summary

- add `MAKER_QUOTE_LIMITED` and `QUOTE_VALIDATION_TIMEOUT_INTERNAL` to the RFQ error enum
- treat unknown future RFQ error codes as correlated operation failures instead of malformed websocket frames
- propagate genuine terminal websocket failures through the RFQ iterator, including close code and reason
- add schema and session regression coverage plus a patch changeset

## Context

A server error code added after the published SDK caused strict frame parsing to fail. The SDK then ended the event iterator and closed an otherwise healthy quoter session. Error-code deployment must not be coupled to an SDK release.

Known codes remain typed and semantically distinct. In particular, `MAKER_QUOTE_LIMITED` is not collapsed into `RATE_LIMITED`. Unknown future codes produce the existing typed operation rejection with `code` omitted, preserving the session and the server-provided message and error ID.

## Verification

Using Node `v24.16.0`:

- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` (228 tests)
- RFQ integration regressions added for the credential-backed CI suite

## Related PRs

- Gateway timeout classification: https://github.com/Polymarket/combos-rfq/pull/176
- Python SDK forward compatibility: https://github.com/Polymarket/py-sdk/pull/161

## Rollout

Publish this SDK so makers can receive the distinct `MAKER_QUOTE_LIMITED` response without terminating their sessions.

---

> [!NOTE]
> **Medium Risk**
> Changes RFQ WebSocket session lifecycle and error typing that makers depend on for quoting; behavior shifts from session-killing parse failures to operation-level rejections and explicit connection-loss errors.
> 
> **Overview**
> **RFQ quoter sessions** are hardened so server-side error and disconnect behavior no longer tears down healthy maker sessions when the SDK lags the gateway.
> 
> **Forward-compatible RFQ errors:** Known codes move to `RfqKnownErrorCode` (including `MAKER_QUOTE_LIMITED` and `QUOTE_VALIDATION_TIMEOUT_INTERNAL`). `RfqErrorCode` is now an open string union so inbound `RFQ_ERROR` frames with unknown `code` values still parse. Correlated quote/cancel/confirmation failures surface as the existing typed rejections with the server `code` and message; the **session and event iterator stay open**.
> 
[...]

PR #206, opened 2026-07-16
fix(rfq): keep internal timeout code private

shazam8253:
## Summary

- remove `QUOTE_VALIDATION_TIMEOUT_INTERNAL` from the known public RFQ error enum
- keep `MAKER_QUOTE_LIMITED` explicitly typed
- preserve forward-compatible handling for unknown server codes without closing the RFQ session

## Context

The gateway maps its internal quote-validation timeout to the public
`SERVICE_UNAVAILABLE` code in Polymarket/combos-rfq#176. The internal
classification should therefore not become part of the SDK's supported public
protocol surface.

This follows up on #198. An older gateway instance can still emit the internal
string during rollout; the open `RfqErrorCode` type continues to pass it
through as an unknown correlated operation error.

## Verification

- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` (241 tests)
- RFQ client integration regressions for known and unknown quote errors

### DIFF
--- a/.changeset/clean-rfq-internal-code.md
+++ b/.changeset/clean-rfq-internal-code.md
@@ -1,0 +1,8 @@
+---
+"@polymarket/bindings": patch
+"@polymarket/client": patch
+---
+
+Keep internal RFQ timeout classifications out of the known public error-code
+enum. Unknown server codes remain forward compatible and continue to surface on
+correlated RFQ operation errors without closing the session.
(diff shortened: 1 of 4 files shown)


cesarenaldi:
Didn't see this one, already addressed in https://github.com/Polymarket/ts-sdk/pull/207.

PR #208, opened 2026-07-17
feat(bindings): add conditionId alias to CLOB shapes, deprecate market

naruto11eth:
The CLOB order book, open order, trade, and builder trade shapes expose a `market` field that actually holds a CTF condition id. This adds a same-valued `conditionId` alias on each and marks `market` as `@deprecated`, so the field name matches what it carries and stays consistent across the SDK.

Additive and non-breaking: both fields are emitted with the same value. The bare `market: z.string()` fields on open order, trade, and builder trade are also upgraded to `CtfConditionIdSchema` (a blind transform, no new validation). Compiler-visible `@deprecated` sits on `OrderBook.market`; the other three carry the additive alias.

Verified: `biome`, `typecheck`, bindings + client tests, and runtime round-trips confirming both keys emit equal values.

---

> [!NOTE]
> **Low Risk**
> Additive, backward-compatible response shaping and typings; no runtime behavior change beyond duplicating an existing field and stricter nominal typing.
> 
> **Overview**
> CLOB **order book**, **open order**, **trade**, and **builder trade** objects now expose **`conditionId`** alongside **`market`**, both set to the same CTF condition id. **`market`** is marked **`@deprecated`** in favor of **`conditionId`** so naming matches the rest of the SDK.
> 
> Zod parsers populate **`conditionId`** from the API’s **`market`** field. **`OpenOrder`**, **`ClobTrade`**, and **`BuilderTrade`** now validate **`market`** with **`CtfConditionIdSchema`** instead of a plain string. Types are declared explicitly and tied to schemas with **`satisfies z.ZodType<…>`** rather than **`z.infer`**. Patch changesets cover **`@polymarket/bindings`** and **`@polymarket/client`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b718fac86a2f5f4b80db29d8d311cb766c98c5cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bindings/src/clob/builder.ts
cesarenaldi: In order to have a `@deprecated` notation on `market`, we need to build the type first and then ensure that the schema satisfies the type.

cesarenaldi: It's a good opportunity to have a type with documented fields.

PR #226, opened 2026-07-20
feat(client): paginate Perps fills with the native cursor and sort

kartojal:
Resolves [DEV-427](https://linear.app/polymarket/issue/DEV-427/support-native-fills-cursorsort-in-ts-and-python-sdks)

## Changes

- `listPerpsFills` / `session.listFills` now forward the API's native `cursor` param instead of the SDK-owned opaque cursor (timestamp window + seen keys), which has been removed for fills. Cursors are treated as opaque pass-through values; `nextCursor` comes from the last fill of each page per the documented cursor protocol.
- New `sort` option (`PerpsSortDirection.Descending` default / `.Ascending`) forwarded as the native `sort` param.
- `ListPerpsFillsRequest` is now a flat `{ start?, end?, sort?, cursor? }` type; the fills-specific default 1-day window is gone, so the API's own defaults apply. Other account history endpoints are unchanged.
- Tests: native fills coverage for first page, cursor continuation, sort forwarding, and exact query serialization; the shared descending-history dedupe tests now run against funding payments.

Note: previously issued SDK-encoded fills cursors are no longer decoded; cursors are short-lived pagination state, so this is not treated as a breaking migration.

## Usage

```ts
const pages = session.listFills({ sort: PerpsSortDirection.Ascending });

const first = await pages.firstPage();
for await (const page of pages.from(first.nextCursor)) {
  // page.items: PerpsAccountFill[]
}
```

---

> [!NOTE]
> **Medium Risk**
> Fills pagination behavior and request defaults change (no implicit 1-day window; old SDK cursors invalid), which can affect integrators mid-pagination, though scope is limited to experimental Perps account reads.
> 
> **Overview**
> **Perps fills listing** now uses the REST API’s native `cursor` and optional `sort` instead of SDK-encoded pagination (timestamp windows, seen keys, and dedupe). `ListPerpsFillsRequest` is a single flat shape `{ start?, end?, sort?, cursor? }`; the client forwards filters on every page and sets `nextCursor` from the last fill’s `tradeId` when `more` is true. Previously issued SDK fills cursors are no longer decoded.
> 
> Bindings add **`PerpsSortDirection`** (`asc` / `desc`) and export it from `@polymarket/client`. Session docs for `listFills` describe default newest-first ordering and opaque API cursors.
> 
> **Tests** cover native fills paging, sort/cursor query strings, and move descending-history dedupe scenarios to **funding payments** (fills no longer use that path). Invalid-cursor coverage uses `listFundingPayments` instead of `listFills`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5899c9c1bd30488cc6f23d077f21bb83445fa90e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/websockets/perps/session.test.ts
cesarenaldi, line 1: Some of these tests still carry some of the details of the previous implementation, where we had to pretend that there was a cursor-based pagination on the client, and we have to fix some shortcomings of doing that with underlying offset based pagination.

I reckon we can simplify the tests here.

packages/client/src/websockets/perps/session.ts
```diff
@@ -355,7 +355,11 @@ export class PerpsSession implements AsyncIterable<PerpsSessionEvent> {
   }
 
   /**
-   * Lists Perps fills with SDK-owned pagination.
+   * Lists Perps fills with cursor pagination.
```
cesarenaldi: nit: even the original wording was off. We don't have to say "cursor pagination." All pagination from the SDKs is cursor-based pagination.

packages/client/src/websockets/perps/actions/account.ts
```diff
@@ -240,86 +241,52 @@ export async function fetchPerpsOrders(
...
-) satisfies z.ZodType<
-  Exclude<ListPerpsFillsRequest, { cursor: PaginationCursor }>
->;
-
-const ListPerpsFillsCursorRequestSchema = z.object({
-  cursor: PaginationCursorSchema,
-}) satisfies z.ZodType<
-  Extract<ListPerpsFillsRequest, { cursor: PaginationCursor }>
->;
+const ListPerpsFillsRequestSchema = PerpsHistoryRequestBaseSchema.extend({
```
cesarenaldi: We no longer need to extend if we don't have a union. You can inline PerpsHistoryRequestBaseSchema here.

kartojal: inlined PerpsHistoryRequestBaseSchema

mcamara (approved):

PR #241, opened 2026-07-24
fix(client): cap pageSize per gamma endpoint and tolerate clamped pagination probes

cesarenaldi:
Follow-up to #240: applies the same silent-truncation fix to the Gamma-backed offset-probe methods (DEV-475).

## Problem

Every Gamma list endpoint silently clamps `?limit` at the query layer; past-cap limits never 400. Same failure mode as #240: a `pageSize` at or above the cap made the `pageSize + 1` probe come back clamped, the SDK saw a full page, and pagination stopped after the first page without an error. Unlike the data service there is no strict-400 change planned upstream, and the caps have already drifted once (markets-tooling #1501 lowered them in May 2026), so clamp tolerance matters here long-term.

## Changes

- Cap `pageSize` in the request schemas one below each endpoint's server-side limit cap (probe headroom), rejecting invalid values with `UserInputError`:
  - `listTags`, `listComments`, `listCommentsByUserAddress`, `listTeams`, `listMarketClarifications`: 99 (server cap 100)
  - `listSeries`: 49 (server cap 50) — `pageSize: 50` truncated silently before this change
- Make the `hasMore` probe clamp-tolerant on these six call sites (`>=` instead of `>`), matching #240. Costs one extra empty-page request when the total is an exact multiple of `pageSize`.

Caps verified against the Gamma limit clamp helper and per-endpoint constants on current markets-tooling `main` (`model.AppendArgs`, `model/constants.go`); the by-user comments route clamps at the same 100 cap as `/comments`. `listMarkets` / `listEvents` are cursor-based and unaffected; search is page-based and out of scope.

Independent of #240 (no overlapping files); mergeable in either order.

## Verification

- `pnpm lint`, `pnpm typecheck`, `pnpm test` (244 unit tests)
- Live integration: tags, comments, series, teams, market-clarifications suites pass; the series and teams walks now page far past the old truncation point

---

> [!NOTE]
> **Medium Risk**
> Changes pagination behavior across six public list methods; fixes data completeness but alters `pageSize` validation (breaking for callers passing 100+) and may add one redundant request per full walk.
> 
> **Overview**
> Fixes **silent first-page-only pagination** on the remaining Gamma offset-paginated list APIs when upstream silently clamps `limit` (same pattern as the earlier data-service fix).
> 
> **`pageSize` validation** now enforces per-endpoint maximums one below the server cap so the `pageSize + 1` probe can work: **99** for `listTags`, `listComments`, `listCommentsByUserAddress`, `listTeams`, and `listMarketClarifications` (cap 100), and **49** for `listSeries` (cap 50). Values above those limits fail at input with **`UserInputError`** instead of truncating quietly.
> 
> **`hasMore` detection** on those six call sites switches from `length > pageSize` to **`length >= pageSize`**, so a full page still advances when the extra probe row is clamped away. Totals that are an exact multiple of `pageSize` may incur **one extra empty request** at the end.
> 
> Integration tests use the new caps (`99` / `49`) for pagination walks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b7e03ff11a0470077d3d1b4fc1db0111b7f2f49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/dev-475-gamma-page-size-caps.md
+++ b/.changeset/dev-475-gamma-page-size-caps.md
@@ -1,0 +1,5 @@
+---
+"@polymarket/client": patch
+---
+
+Fix silent pagination truncation on the remaining offset-paginated list methods, matching the fix already applied to data-backed methods. The upstream service silently clamps `limit` per endpoint, so a `pageSize` at or above the cap made the look-ahead probe come back clamped and pagination stopped after the first page without an error. `pageSize` is now validated against each endpoint's cap (`listTags`, `listComments`, `listCommentsByUserAddress`, `listTeams`, `listMarketClarifications` 99; `listSeries` 49) and rejects invalid values with `UserInputError`. The `hasMore` probe on these methods is also clamp-tolerant now: a full page continues pagination instead of relying on the extra probe row, at the cost of one extra empty-page request when the total count is an exact multiple of `pageSize`.
--- a/packages/client/src/actions/comments.ts
+++ b/packages/client/src/actions/comments.ts
@@ -34,7 +34,8 @@ import { snakeCase, toSearchParams } from './params';
 const ListCommentsRequestSchema = z.object({
   ascending: z.boolean().optional(),
   cursor: PaginationCursorSchema.optional(),
-  pageSize: PageSizeSchema.default(20),
+  // Upstream caps limit at 100 and the probe requests pageSize + 1.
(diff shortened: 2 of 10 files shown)


cesarenaldi:
Consolidated into #240.

PR #250, opened 2026-07-29
refactor(client): lift Exchange v3 order construction into the exchange module

kartojal:
Prerequisite refactor for DEV-424. Moves the Exchange v3 domain helper, maker/taker amount math, salt generation, and side encoding from `websockets/rfq/signing.ts` into `src/exchange.ts`, leaving the RFQ signing module as a thin call site. Pure move-and-rename; no behavior change, maker-side RFQ tests stay green.

Linear: DEV-424

---

> [!NOTE]
> **Low Risk**
> Pure move-and-rename refactor; order math and signing flow are unchanged and RFQ tests are expected to stay green.
> 
> **Overview**
> **Moves Exchange v3 order-building helpers** from `websockets/rfq/signing.ts` into `exchange.ts` so RFQ signing stays a thin orchestration layer (DEV-424 prerequisite).
> 
> New `@internal` exports in `exchange.ts`: `createExchangeV3OrderDomain`, `calculateExchangeOrderMakerAmount`, `calculateExchangeOrderTakerAmount`, `generateExchangeOrderSalt`, and a public `encodeExchangeOrderSide` (previously private). `signRfqQuoteOrder` now imports these instead of local duplicates; the old private helpers (`createRfqExchangeDomain`, `makerAmount`, `takerAmount`, etc.) are removed.
> 
> Documented as a patch-level internal refactor with **no intended behavior change**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91b5c7d129178f2dd8f72fa49333ae7a256bb935. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #251, opened 2026-07-29
feat(client): requester-side combo RFQ support (request quote, accept, wait for fill)

kartojal:
Implements DEV-424. Adds `client.requestComboQuote(...)`, `client.acceptComboQuote(...)`, and `client.waitForComboFill(...)` on SecureClient, plus the unbound `fetchRfqStatus` action, over a new builder-gateway ServiceClient (L2 + builder auth headers, builder headers omitted on status reads). Wire schemas and enums live in `@polymarket/bindings/combos`; rejections throw the new `RfqRequestRejectedError` with a classified `RfqRejectionCode` (unknown codes fall back to `INVALID_RFQ`, original error on `cause`).

Notes vs the ticket spec, verified against combos-rfq main:
- Accept is a plain held request (SSE removed); if the gateway still reports `AWAITING_MAKER_CONFIRMATION` after its hold window the SDK resumes over the status endpoint, so `TimeoutError` was added to `AcceptComboQuoteError`.
- `acceptComboQuote` params carry the winning quote data (amounts, position ID, builder code) because the status endpoint never returns the bundle; `requestComboQuoteResult` is passed to it directly.
- `ComboQuote` amounts are decimal strings (exact e6 round-trip for order signing).
- `WaitForComboFillError` additionally includes `UserInputError` (params are Zod-validated like the other methods) and `RfqRequestRejectedError` (status reads can be rejected, e.g. for an unknown or not-yet-accepted RFQ); both genuinely propagate through the wait loop.

Stacked on #250.

Linear: DEV-424

---

> [!NOTE]
> **High Risk**
> Touches trading and order-signing paths plus builder authentication against a new gateway, so bugs can cause incorrect accepts, failed fills, or auth misuse with real funds.
> 
> **Overview**
> Adds **requester-side combo RFQ** support so builders can request a quote, accept it with a signed order, and wait for fill over a new builder-gateway client.
> 
> Exposes `requestComboQuote`, `acceptComboQuote`, and `waitForComboFill` on the secure client (plus unbound `fetchRfqStatus`). Winning quotes are self-contained JSON (including SELL `netReceive`) and can cross process boundaries before acceptance. Business outcomes like no quotes, maker decline, or an expired window return as results; gateway rejections throw `RfqRequestRejectedError` with a classified `RfqRejectionCode`.
> 
> Also wires builder-gateway auth (builder headers on mutating calls only), Zod schemas/enums in `@polymarket/bindings/combos`, and propagates JSON error `code` on `RequestRejectedError`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c94d32d6b0f442534cb5263e9b8883765338792a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cesarenaldi:
`acceptComboQuote` currently depends on signing inputs stored in a module-level `WeakMap` for the same `SecureClient` instance. This breaks a normal omnibus integration where requesting and accepting are separate user interactions:

```text
[Browser]
    |
    | 1. Request quote
    v
[Worker A / SecureClient A]
    |
    | POST /v1/builder/rfq/requests
    v
[Builder Gateway]
    |
    | RFQ + quote + signing inputs
    v
[Worker A / SecureClient A]
    |
    | Store signing inputs in WeakMap(Client A)
    | Return public quote
    v
[Browser]
    |
    | 2. Accept { rfqId, quoteId }
    v
[Worker B / SecureClient B]
    |
    | Look up signing inputs in WeakMap(Client B)
    v
[CACHE MISS -> UserInputError]
    |
    X Builder Gateway is never called
```

Worker B could also be Worker A after a restart or client recreation. Requiring the same process would force omnibus integrators to use sticky sessions or maintain another application-owned cache.

Can we make `ComboQuote` self-contained and serializable, then pass it directly to `acceptComboQuote`? For example:

```ts
export type ComboQuote = {
  rfqId: RfqId;
  quoteId: RfqQuoteId;
  builderCode: BuilderCode;
  direction: OrderSide;
  positionId: PositionId;
  makerAmount: DecimalString;
  takerAmount: DecimalString;
  blendedPrice: DecimalString;
  totalRequired: DecimalString;
  expiresAt: EpochMilliseconds;
};

const result = await client.requestComboQuote(...);

if (result.quote !== null) {
  await client.acceptComboQuote(result.quote);
}
```

These fields are JSON-serializable and should be treated as a public contract so an integrator can validate, persist, or route the quote with the user action. This keeps one acceptance path while removing the hidden cache and its 100-entry eviction limit.

packages/client/src/actions/rfq.ts
```diff
@@ -361,3 +417,895 @@ export async function openRfqSession(
...
+        { json: request, timeout: ACCEPT_QUOTE_TIMEOUT_MS },
+      )
+      .andThen(validateWith(BuilderRfqStatusResponseSchema));
+  }
+
+  let accepted = await postAcceptance();
+
+  // Acceptance is idempotent server-side, so a single retry safely covers a
+  // connection dropped while the request is held through maker last look.
+  if (accepted.isErr() && accepted.error instanceof TransportError) {
```
mcamara: why is `TransportError` the only error being retried? `UnexpectedResponseError` could also be retried, right?

kartojal: yes, added it

mcamara (approved):

PR #255, opened 2026-07-31
feat(bindings): type notification payloads per notification kind

kartojal:
Types `Notification` per DEV-459: `NotificationType` enum (all 10 wire kinds), branded `ApiKey` owner, and a discriminated union tying each `payload` shape to its notification kind. Payload shapes verified against the producing services.

Linear: DEV-459

---

> [!NOTE]
> **Medium Risk**
> Breaking type changes for `Notification` consumers and stricter parsing can reject previously accepted malformed API data or fail entire notification fetches when one known notification is bad.
> 
> **Overview**
> **Typed account notifications** replace the loose `Notification` shape: a `NotificationType` enum (10 wire kinds), `owner` as branded `ApiKey`, and a discriminated union so narrowing on `type` narrows `payload` (order lifecycle, market lifecycle, payouts, comments, auto-redeem, etc.). Schemas live in a new `notifications` module; wire fields are normalized to camelCase in transforms.
> 
> **`fetchNotifications` parsing** drops notification kinds the SDK does not know about, but **fails the whole response** if a known kind’s payload does not validate. Client docs and `NotificationType` export reflect that.
> 
> **Shared Zod parsers** for combo condition IDs, EVM addresses, question IDs, and transaction hashes now **refine before transform**, so bad values surface as schema errors instead of throws from `expect*` helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42bf0d69ea21ee9858d12452f9b73b3621c9fa85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/bindings/src/clob/notifications.ts
```diff
@@ -0,0 +1,533 @@
...
+>;
+
+/**
+ * One outcome token inside a market lifecycle notification payload. On a
+ * market-resolved notification, `winner` marks the winning outcome.
+ */
+export const MarketNotificationTokenSchema = z
+  .object({
+    outcome: z.string(),
+    price: z.number().optional(),
```
mcamara: shouldn't it be `DecimalString`? same in line 152, 263, 278, 321 and 349

mcamara (approved):
small nitpick, the pr looks good otherwise

Issue #253, opened 2026-07-30
fetchLastTradePrice throws on the CLOB no-trade sentinel (side: "")


jpolym:
For a token that has never printed a trade, the CLOB single read returns a sentinel and the batch read omits the token:

```
$ curl 'https://clob.polymarket.com/last-trade-price?token_id=100786328878817888672127899346115474991295835713002536260414835168071794216281'
{"price":"0.5","side":""}

$ curl -X POST 'https://clob.polymarket.com/last-trades-prices' -H 'Content-Type: application/json' -d '[{"token_id":"100786328878817888672127899346115474991295835713002536260414835168071794216281"},{"token_id":"72427307104941456363245557419519246714566250320546964438571016095329563907082"}]'
[{"price":"0.999","side":"BUY","token_id":"72427307104941456363245557419519246714566250320546964438571016095329563907082"}]
```

Captured 2026-07-30 on market uel-ben-stg-2026-07-30-spread-away-2pt5, but any never-traded token reproduces: the server hardcodes price "0.5" / side "" when no trade exists (clob-v2 ledger GetLastTradePrice, in place since May 2024).

LastTradePriceSchema (packages/bindings/src/clob/market-data.ts) pins side to the BUY|SELL enum, so fetchLastTradePrice throws a ZodError on every never-traded token — same-day sports markets hit this routinely. It failed our production smoke (e2e-defi run 30544322966) on canary-20260730090047, but the gap exists in every version.

Suggested fix: model "no trade yet" as an explicit discriminated result rather than silently accepting side "" — the sentinel price "0.5" is indistinguishable from a real 0.5 trade unless the caller inspects side. E.g. `{ hasTraded: false } | { hasTraded: true; price; side }`, or side: null with the placeholder price documented. Also worth documenting that /last-trades-prices omits never-traded tokens, so callers can't assume a row per requested token.

kartojal:
@jpolym thanks for reporting this, we will have a look and propose a solution

kartojal:
@jpolym fixed in https://github.com/Polymarket/ts-sdk/pull/279 and released in version v0.6.0.

Now `fetchLastTradePrice()` returns `null` if there is no trades in a token.

PR #279, opened 2026-08-07
fix(client): handle tokens without trades

kartojal:
Returns a nullable side for tokens without trades and documents the placeholder price response.

DEV-506
https://github.com/Polymarket/ts-sdk/issues/253

---

> [!NOTE]
> **Medium Risk**
> Public API return type changes for last-trade reads; callers that assumed a non-null object may need null checks, though behavior now matches the API.
> 
> **Overview**
> **`fetchLastTradePrice`** now resolves to **`null`** when the CLOB returns a placeholder response (price with an empty `side`), instead of treating that as a valid trade.
> 
> The **`LastTradePriceSchema`** in `@polymarket/bindings` maps `side: ''` to `null` and still rejects invalid sides. **`@polymarket/client`** updates the action and **`DataActions`** return type to **`Promise<LastTradePrice | null>`**, with docs noting that **`fetchLastTradePrices`** omits untraded tokens and results must be matched by **`tokenId`**, not request order.
> 
> Tests cover schema parsing and public type expectations; the market-prices example uses optional chaining for display.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45e00a23541e1e8c7dd94a8486abb24f839f22ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/decorators/data.ts
```diff
@@ -169,6 +169,9 @@ export type DataActions = {
   /**
    * Fetches the last traded price for a token.
    *
+   * When the token has not traded, `side` is `null` and `price` is the `"0.5"`
```
cesarenaldi: What is this synthetic 0.5? could it be misleading?

kartojal: as brunson points, this is the default response from the API service if there is no trades, we can:
- Keep it 0.5 so is equal as the API service
- Return lastPrice 0 and side null
- Return NoTraded error

open for your feedback here

cesarenaldi: If we are saying that `side=null` (and consequently `price=0.5`) means "no information available", I think we should codify the no information available as `null` of the entire response.

I would not throw cause it's not a caller mistake, nor a wrong state of the system. There is a genuine time window where someone asking for a trade price on a new market can yield no value.

mcamara (approved):

PR #280, opened 2026-08-07
feat(bindings): support PolyV2 market discovery

cesarenaldi:
## Summary

First PR in the DEV-544 PolyV2 unified TypeScript SDK stack.

- Introduce `ConditionId` as the protocol-neutral ordinary-market condition type while retaining `CtfConditionId` as a deprecated compatibility alias.
- Expose required-nullable `positionId` alongside `tokenId` on normalized market outcomes.
- Preserve the market-level `positionIds` array as a deprecated compatibility surface.
- Parse CTF and PolyV2 condition IDs from market responses without inferring protocol version from their byte layout.
- Migrate internal CLOB, Data, subscriptions, Combos, caches, and call builders to the generalized condition type.

## Why

Market discovery is the first layer of PolyV2 SDK support. It gives callers outcome-level position IDs and establishes the shared condition identifier model needed by the subsequent CLOB order-placement, CLOB normalization, and position-lifecycle PRs.

## Compatibility

- Existing outcome `tokenId` fields remain unchanged.
- Existing market-level `positionIds` remains available and is marked deprecated.
- Existing consumers of the public `CtfConditionId` type remain source-compatible through the deprecated alias.
- No upstream wire field names change.

## Validation

- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test` — 91 bindings tests and 295 client tests

Linear: [DEV-544](https://linear.app/polymarket/issue/DEV-544/polyv2-define-and-implement-unified-typescript-sdk-support)

---

> [!NOTE]
> **Medium Risk**
> Wide public API surface rename with deprecated aliases preserves source compatibility, but consumers typing against CtfConditionId should migrate; new outcome fields and looser gamma condition parsing could affect discovery code paths.
> 
> **Overview**
> Introduces **PolyV2 market discovery** in bindings and client: normalized binary outcomes now expose nullable **`positionId`** next to **`tokenId`** (legacy CLOB-only, PolyV2-only, or both), while market-level **`positionIds`** stays available but is **deprecated** in favor of per-outcome IDs.
> 
> Renames the shared market identifier from **`CtfConditionId`** to protocol-neutral **`ConditionId`** (`ConditionIdSchema`, `toConditionId`, `OptionalConditionIdSchema`) across CLOB, data, gamma, subscriptions, combos, and client actions/caches/ABIs. **`CtfConditionId*`** exports remain as deprecated aliases with identical runtime behavior. Gamma parsing adds **`ConditionIdResponseSchema`** (hex-only, no byte-length gate) so CTF and PolyV2 condition IDs from API responses are accepted without inferring protocol from layout.
> 
> Tests cover outcome identifier combinations, polymorphic condition IDs, and compatibility aliases. Minor version bump for `@polymarket/bindings` and `@polymarket/client`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1384740923b3aa8ae78fe32524e3c23e1cf113c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cesarenaldi:
@brunson-bot Thanks — addressed the actionable parts in afd86aa.

- The response/input schema difference is intentional. At the Gamma response boundary we require usable hex but do not infer protocol support from byte length; SDK-owned operation inputs retain the known 31/32-byte validation. I added a comment on `ConditionIdResponseSchema` to make that boundary decision explicit.
- Added the negative `MarketSchema.safeParse` regression case so a non-hex condition ID is reported as a validation failure rather than escaping as a raw `TypeError`.
- Tightened the compatibility wording to say that consumers of the public `CtfConditionId` type remain source-compatible; the bindings-only runtime schemas/converter are not part of that commitment.

I left the response schema length policy unchanged. A condition ID discovered from Gamma can be represented for discovery without claiming that every future byte layout is already supported as an SDK operation input.

I also did not add the live `positionId` assertion: the live endpoint cannot reliably provide a representative CTF-with-Combos market, while the normalization fixture covers undeployed, CTF-only, PolyV2-only, and combined shapes. The mapping intentionally follows Gamma's parallel outcome arrays by index.

kartojal (approved):

PR #288, opened 2026-08-10
feat(client): expose trading approval state (DEV-564)

### LINKED ISSUE #286: Public API to read trading-approvals state without triggering the signing workflow
`setupTradingApprovals()` already does the right thing internally — `resolveMissingTradingApprovals` reads the on-chain allowances and the workflow only signs/sends what's missing. But that resolver isn't exported, so there's no way to just *read* the state: e.g. to check on connect whether any approvals are missing and show the user a setup popup first (the way polymarket.com does), without ever prompting a signature.

The low-level pieces (`erc20AllowanceCall`, `erc1155IsApprovedForAllCall`, decoders, `client.rpc.ethCallBatch`) are public and work on a plain public client, but the canonical requirement list in `getRequiredTradingApprovals` is not exported — so hand-rolling the check means copying a list that can drift (the docs' manual walkthrough currently lists 4 approvals while the SDK enforces 15).

Ask: something like `getTradingApprovalsState(client)` → `{ missing, isFullyApproved }` — a plain read, no signer required.

kartojal:
## Summary
- add signer-free trading approval state reads for public clients
- add authenticated-wallet convenience reads for secure clients
- reuse the canonical missing-approval resolver in setupTradingApprovals
- add read-only integration coverage and a client package changeset

Closes #286
Linear: DEV-564

## Testing
- pnpm lint
- pnpm typecheck
- client unit suite: 295 passed
- approvals integration suite: 2 passed, 5 credential-dependent skipped

---

> [!NOTE]
> **Low Risk**
> Mostly additive read APIs and stricter input validation on secure account reads; setup approvals behavior is aligned with the shared resolver rather than changed semantically.
> 
> **Overview**
> Adds **read-only trading approval state** via `fetchTradingApprovalsState`: public clients pass a `user` address; secure clients get a no-arg helper that inspects the authenticated wallet. Results include `isFullyApproved` and structured `missing` ERC-20 / ERC-1155 requirements, with the related types exported on the wallet decorator.
> 
> **`setupTradingApprovals`** now shares the same missing-approval resolver, parameterized by wallet and usable on **`BaseClient`** (RPC-only). Malformed `eth_call` decode failures surface as **`UnexpectedResponseError`** instead of bubbling raw decode errors.
> 
> **Public clients** gain **`PublicWalletActions`** (`fetchTradingApprovalsState` only) through `allActions`. Secure **account** helpers stop treating invalid request shapes (`null`, arrays, primitives, `user: null`) as “default my wallet”—those inputs fail with **`UserInputError`** at validation time.
> 
> Integration tests cover signer-free `eth_call`-only reads and secure wallet defaulting; portfolio tests assert invalid secure `listPositions` input is rejected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df294a140e57f3bba6eef3afb25bd3c6183b8ef2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/decorators/default-wallet.ts
cesarenaldi, line 1: As discussed this approach was evaluated and not adopted cause, although it produce valid TS code, it does produce a more cryptic IDE suggestions.

packages/client/src/decorators/wallet.ts
```diff
@@ -15,12 +18,68 @@ import {
...
+   *   user: '0x1234…',
+   * });
+   * ```
+   */
+  fetchTradingApprovalsState(
+    request: FetchTradingApprovalsStateRequest,
+  ): Promise<TradingApprovalsState>;
+};
+
+/** Parameters for reading trading approvals with a secure client. */
```
cesarenaldi: The arbitrary-wallet read makes sense on `PublicClient`, but `SecureClient` should stay bound to `client.account.wallet`. Since `user` is the only request field, can we make `secureClient.fetchTradingApprovalsState()` parameterless and always inspect the authenticated account? This keeps the natural preflight and `setupTradingApprovals()` workflow aligned to the same wallet.

cesarenaldi (approved):

PR #292, opened 2026-08-14
feat(perps): expose server time

kartojal:
[DEV-466](https://linear.app/polymarket/issue/DEV-466/plan-integrate-server-time-from-v1infotime-into-perps-request): Exposes the Perps server clock.

Tests: lint, typecheck, build, unit PASS; public live PASS.

---

> [!NOTE]
> **Low Risk**
> Additive, read-only public API following existing Perps info patterns; no auth, signing, or request-timestamp behavior changes.
> 
> **Overview**
> Adds a **public Perps clock** so callers can read the exchange server time in epoch milliseconds via `GET /v1/info/time`.
> 
> **Bindings** introduce `GetServerTimeResponseSchema` (integer ms only). **Client** adds `getServerTime` on the low-level actions and `client.getServerTime()` on `PublicPerpsActions`, returning branded `EpochMilliseconds` with the usual transport/rate-limit error guard. Docs stress this is a read-only probe and **does not** change how the SDK timestamps or signs Perps requests today.
> 
> Coverage includes schema tests, type-level export checks, and a live integration assertion that server time is within ±60s of local wall clock. Minor version bumps are recorded for `@polymarket/bindings` and `@polymarket/client`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a2fe012aa50d3bbaf0ac13d27adec66c450f350. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/actions/perps.ts
+++ b/packages/client/src/actions/perps.ts
@@ -1,4 +1,5 @@
 import {
+  type EpochMilliseconds,
   type PaginationCursor,
   PaginationCursorSchema,
   toPaginationCursor,
@@ -12,6 +13,7 @@ import {
   FetchPerpsStatisticsResponseSchema,
   FetchPerpsTickersResponseSchema,
   FetchPerpsTradesResponseSchema,
+  GetServerTimeResponseSchema,
   type PerpsBook,
   PerpsBookSchema,
   type PerpsCandle,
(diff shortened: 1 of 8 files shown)


kartojal:
Naming and scope are acceptance decisions recorded on DEV-466: the ticket explicitly requires `getServerTime`, and the paired DEV-466/DEV-467 clock decision deliberately ships only the additive server-clock read while deferring signed-request clock policy. I removed the redundant mocked happy-path client test; binding tests retain response validation and the public integration test proves the real endpoint.

packages/client/src/actions/perps.ts
```diff
@@ -793,6 +795,48 @@ export async function fetchPerpsFees(
...
+ * @remarks
+ * This is a low-level function. Most SDK consumers should prefer the client instance API.
+ * This read does not change how the SDK timestamps or signs Perps requests.
+ *
+ * @throws {@link GetServerTimeError}
+ * Thrown on failure.
+ *
+ * @experimental This API may change in a breaking way in any release, including patch releases.
+ */
+export async function getServerTime(
```
cesarenaldi, line 828: What developer workflow is this standalone read intended to solve? DEV-466 starts by asking where server time is needed and whether the SDK should abstract clock synchronization, but this PR leaves signed requests on `Date.now()` and gives callers no way to feed the returned value into signing. If server time is needed for correctness, I would expect the SDK to own the offset and synchronization rather than require callers to fetch it. If there is an independent use case for exposing the raw read, can we document it? As a smaller naming point, `getServerTime` is generic on the flat client and does not follow the existing `fetchPerps*` convention.

osr21:
For the standalone clock probe, a small usage note could prevent consumers from estimating skew incorrectly. The useful offset is based on the request midpoint, not the response time:

`t0 = Date.now(); server = await client.getServerTime(); t1 = Date.now();`

`offsetMs = server - Math.floor((t0 + t1) / 2)`
`uncertaintyMs = Math.ceil((t1 - t0) / 2)`

Callers should discard/highlight samples with large RTT, keep the lowest-RTT recent sample, and report both offset and uncertainty. Computing `server - t1` systematically folds network latency into the apparent clock skew.

I agree with the existing discussion that signing should eventually consume an SDK-owned offset rather than exposing timestamp overrides on every command. Until that follow-up lands, documenting the midpoint pattern gives this read-only primitive a concrete and correct observability use case: warn that the local clock is outside tolerance before starting a delegated Perps session, without implying that the current SDK corrects signatures.

PR #295, opened 2026-08-14
feat(perps): add batch leverage updates

kartojal:
[DEV-552](https://linear.app/polymarket/issue/DEV-552/unified-sdks-lack-batch-perps-leverage-updates): Adds ordered batch Perps leverage updates with mixed-result handling.

---

> [!NOTE]
> **Medium Risk**
> New trading-session API changes leverage and margin mode on live accounts with non-atomic batch semantics; incorrect client handling of partial failures could leave account config out of sync.
> 
> **Overview**
> Adds **ordered batch Perps leverage updates** via a new `updateLeverages` command and `session.updateLeverages()` (plus `updatePerpsLeverages` in the trading layer). Callers send 1–100 unique instrument updates in one signed WebSocket command; responses are an array aligned with request order, mixing per-instrument `ok` results and `err` rejections (empty errors are preserved; whole-request failures still throw).
> 
> Bindings gain `PerpsUpdateLeverageRejection` and `PerpsUpdateLeverageBatchResult` schemas. Client validation rejects empty batches, duplicates, and invalid fields before the command runs. Docs note updates are **sequential and not atomic**, including guidance on `internal_error` and retries.
> 
> Unit/session tests cover signing, `toPerpsCommandBodyOp`, validation boundaries, and mixed batch outcomes; a metered integration test updates and restores leverage using a one-item batch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c7fdc5ddbfb2bdfdfb5e356506c9957bc3fe493. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/websockets/perps/actions/trading.ts
+++ b/packages/client/src/websockets/perps/actions/trading.ts
@@ -24,10 +24,10 @@ import {
   PerpsTimeInForce,
   PerpsTpSlKind,
   PerpsTpSlScope,
+  type PerpsUpdateLeverageBatchResult,
+  PerpsUpdateLeverageBatchResultSchema,
   type PerpsUpdateLeverageResult,
   PerpsUpdateLeverageResultSchema,
-  type PerpsUpdateLeveragesResult,
-  PerpsUpdateLeveragesResultSchema,
 } from '@polymarket/bindings/perps';
 import type {
   PerpsOrderUpdateEvent,
@@ -1217,7 +1217,9 @@ export const UpdatePerpsLeveragesError = makeErrorGuard(
  * Updates are processed sequentially and are not atomic. Results preserve
  * request order. Per-instrument rejections, including `internal_error`, are
  * returned as data; `internal_error` may represent an unknown application
- * outcome for that instrument.
+ * outcome for that instrument. A whole-request `internal_error` also has an
+ * unknown application outcome; reconcile account state before retrying the
+ * batch.
  *
  * @throws {@link UpdatePerpsLeveragesError}
  * Thrown when the complete request is rejected or cannot be sent.
@@ -1227,7 +1229,7 @@ export const UpdatePerpsLeveragesError = makeErrorGuard(
 export async function updatePerpsLeverages(
   client: PerpsCommandExecutor,
   request: UpdatePerpsLeveragesRequest,
-): Promise<PerpsUpdateLeveragesResult> {
+): Promise<PerpsUpdateLeverageBatchResult[]> {
   const params = parseUserInput(request, UpdatePerpsLeveragesRequestSchema);
   return await client.executeCommand(
     {
@@ -1240,7 +1242,7 @@ export async function updatePerpsLeverages(
         ]),
       ],
     },
-    PerpsUpdateLeveragesResultSchema,
+    z.array(PerpsUpdateLeverageBatchResultSchema),
   );
 }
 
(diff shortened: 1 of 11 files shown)


packages/client/src/websockets/perps/actions/trading.ts
```diff
@@ -1149,6 +1151,101 @@ export async function updatePerpsLeverage(
   );
 }
 
+const MAX_UINT32 = 4_294_967_295;
+const BatchPerpsLeverageUpdateSchema = z.object({
+  instrumentId: PerpsInstrumentIdSchema.refine(
+    (instrumentId) => instrumentId <= MAX_UINT32,
+    'Expected instrumentId to be at most 4294967295',
```
cesarenaldi, line 1158: I don't think this check give much value. Instrument IDs are retrieve from the API, not guessed. At most if this is important should be self-contained in the `PerpsInstrumentIdSchema`.

I would drop this refine.

packages/client/src/websockets/perps/actions/trading.ts
```diff
@@ -1149,6 +1151,101 @@ export async function updatePerpsLeverage(
...
+    .array(BatchPerpsLeverageUpdateSchema)
+    .min(1)
+    .max(100)
+    .superRefine((updates, context) => {
+      const instrumentIds = new Set<number>();
+      for (const [index, update] of updates.entries()) {
+        if (instrumentIds.has(update.instrumentId)) {
+          context.addIssue({
+            code: 'custom',
+            message: 'Expected each instrumentId to be unique',
```
cesarenaldi, line 1175: Is this an hard rule on the BE?

I understand they are updated sequentially by the perps engine.

I am not 100% sure but if some of these are not in isolation mode, there could be a sequence that allows an operator in a single update to shifts leverage across positions without impacting their margin limits.

packages/client/tests/integration/perps.test.ts
```diff
@@ -243,6 +245,98 @@ describe('Perps integration', () => {
     },
   );
 
+  it.runIf(runMeteredTests)(
```
cesarenaldi, line 248: If we end up dropping [this](https://github.com/Polymarket/ts-sdk/pull/295/changes#r3830214149) this whole test could be massively simplified by adjusting the leverage on a single position to a value and back in the same request. As far as integration test goes it would be enough to proof the SDK works with the API.

osr21:
The partial-success shape looks useful for a live trading client. One caller pattern may be worth adding to the TSDoc example: **snapshot first, reconcile selectively, never replay the whole batch**.

Suggested flow:

- Read current account config for every requested instrument and omit no-op updates.
- Submit `updateLeverages` once.
- Accept each explicit `ok`/`err` slot independently.
- For a per-item `internal_error`, re-read only that instrument and compare both leverage and margin mode with the requested target.
- For a whole-request unknown outcome, re-read all requested instruments; classify each as applied/not-applied/indeterminate before allowing any retry.
- If retry is needed, construct a **new batch containing only verified not-applied updates** rather than replaying successful or indeterminate slots.

That distinction is important in order-entry flows: replaying a full sequential, non-atomic batch after an ambiguous acknowledgement can race a position/open-order change and can also overwrite a later user choice. The PR already documents reconciliation; a compact code sample showing selective retry would prevent consumers from treating this like an atomic request.

osr21 (approved):

PR #300, opened 2026-08-19
feat(sdk): add scoped session keys (DEV-588)

kartojal:
Adds authorizeSessionKey and ordinary SecureClient session signing. Uses relayer transaction polling until TRA-354 provides session-key listing.

---

> [!NOTE]
> **High Risk**
> Changes wallet classification and signing envelopes for orders and gasless Deposit Wallet execution, plus on-chain session-key authorize/revoke—core auth and trading paths.
> 
> **Overview**
> Adds **scoped session-key lifecycle** for Deposit Wallets: list active keys from CLOB, authorize via relayer (builder API key + signed `authorizeSessionSigner` batch), and revoke via gasless relayer. Scopes default to **`ALL`** when omitted; known scopes are enumerated in bindings while arbitrary scope strings still pass through.
> 
> Introduces **`SignerType`** (`OWNER` vs `SESSION_KEY`) on account identity. When the signer does not deterministically own the wallet, the client now classifies the account as a Deposit Wallet **session key** (replacing the previous hard error for mismatched addresses). **`wrapDepositWalletSignature`** wraps order and gasless batch signatures into the session-signer envelope; **`OrderSignature`** and **`SessionSignerSignature`** extend signed-order typing.
> 
> **Combos / RFQ** entry points reject session-key clients. Decorator methods **`fetchSessionKeys`**, **`authorizeSessionKey`**, and **`revokeSessionKey`** mirror the new actions. Bindings add CLOB/relayer session-signer schemas and **`SessionSignerKnownScope`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14a11229e9638623d14c2b242edebd292a8bca86. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,322 @@
...
+  | SessionKeyScope.CLOB
+  | SessionKeyScope.RFQ
+  | SessionKeyScope.COMBOSRFQ
+  | SessionKeyScope.BLOCKTRADE;
+
+/** Status of a usable session-key grant. */
+export enum SessionKeyStatus {
+  /** The session signer may use its granted scopes until expiry or revocation. */
+  ACTIVE = 'ACTIVE',
+}
```
cesarenaldi: Is this enum mirroring a BE state? Is it going to grow?

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,322 @@
...
+import { completeWith } from '../workflow';
+import {
+  buildDepositWalletExecuteRequest,
+  GaslessTransactionHandle,
+} from './gasless';
+
+/** Venue authorization attached to a session-key grant. */
+export enum SessionKeyScope {
+  /** Reserved scope value. It cannot be requested. */
+  UNSPECIFIED = 'UNSPECIFIED',
```
cesarenaldi: What is the use case for this value?

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,322 @@
...
+  CLOB = 'CLOB',
+  /** Request-for-quote trading. */
+  RFQ = 'RFQ',
+  /** Combos request-for-quote trading. */
+  COMBOSRFQ = 'COMBOSRFQ',
+  /** Block trading. */
+  BLOCKTRADE = 'BLOCKTRADE',
+}
+
+/** Scope values that may appear in an authorization request or active grant. */
```
cesarenaldi: In active grants, which I assume is the `ActiveSessionKey` type, we have an array of `SessionKeyGrantScope`.

Based on this a value for `ActiveSessionKey.scope` could be:

```
{
   scopes: ["ALL", "COMBOSRFQ", "CLOB"]
}
```

which reads odd. `ALL` seems to not fit into the data model, might just be a shortcut value for the `AuthorizeSessionKeyRequest`.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,322 @@
...
+ *   validUntil: Math.floor(Date.now() / 1_000) + 15 * 60,
+ * });
+ * ```
+ *
+ * @throws {@link AuthorizeSessionKeyError}
+ * Thrown on failure.
+ */
+export function authorizeSessionKey(
+  client: BaseSecureClient,
+): (request: AuthorizeSessionKeyRequest) => Promise<AuthorizeSessionKeyResult> {
```
cesarenaldi: This action function differ from existing pattern.

It uses a curried function while existing actions are normal functions and gets a partial application (via `.bind()` at the time they are bound to the instance (in the decorator files).

packages/client/src/actions/session-keys.ts
kartojal: from the SDK point of view is not needed, dropping it

packages/client/src/actions/session-keys.ts
kartojal: we should keep it due sessions keys could be generated outside of the SDK in an unorthodox way, and we may require to support to read session keys with valid scopes even if they do not make much sense, like your example `[ALL, CLOB, COMBOSRFQ]`

packages/client/src/actions/session-keys.ts
kartojal: agree, added a rule in AGENTS.md to enforce flat client-first action signature by default over currying if not needed.

packages/client/src/actions/session-keys.ts
kartojal: currently is not clear from BE if we will have a single getter to retrieve the state of a single session key and is not clear how that would look like.

could be:
ACTIVE,
EXPIRED,
REVOKED

or just ACTIVE, NOT_ACTIVE (which i could consider just a boolean)

dropping this until we have more details from BE

packages/client/src/actions/session-keys.ts
cesarenaldi: Is `ALL` an actual value from the API?

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+  /** Requested scopes. Defaults to `ALL`, which must appear alone. */
+  scopes?: SessionKeyScope[];
+  /** Absolute expiry as whole future Unix seconds. */
+  validUntil: number;
+};
+
+const SessionKeyScopeSchema = z
+  .string()
+  .min(1)
+  .transform((value): SessionKeyScope => value);
```
cesarenaldi: Intentional. The schema stays open on purpose, so let's record that decision in the code before someone "fixes" it into an enum later.

```suggestion
// Intentionally open: scope names are validated server-side, and the SDK
// must accept scopes it does not enumerate.
const SessionKeyScopeSchema = z
  .string()
  .min(1)
  .transform((value): SessionKeyScope => value);
```

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,117 @@
...
+  sessionSignerAddress: EvmAddress;
+  /** Owner signature authorizing the wallet call. */
+  signature: string;
+  /** Session-signer expiry encoded as whole Unix seconds. */
+  validUntil: string;
+  /** Deposit Wallet that grants the authorization. */
+  walletAddress: EvmAddress;
+};
+
+export const RelayerAuthorizeSessionSignerRequestSchema = z.object({
```
cesarenaldi: Can we drop `RelayerAuthorizeSessionSignerRequestSchema` and `RelayerRevokeSessionSignerRequestSchema`, plus the test covering the former? Nothing parses them. Both actions build the payload as a typed literal and post it, so the plain request types are enough.

FYI on the convention here: schemas that parse SDK input live in `packages/client`, usually alongside the action that owns the input contract (`createAuthorizeSessionKeyRequestSchema` is already doing that job). Bindings schemas are only for representing upstream constraints and shapes we actually parse, i.e. responses and raw wire payloads. There are AGENTS.md guidelines around this split.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+    );
+
+  return schema satisfies z.ZodType<
+    ParsedRevokeSessionKeyRequest,
+    RevokeSessionKeyRequest
+  >;
+}
+
+/** Result of a confirmed session-key revocation. */
+export type RevokeSessionKeyResult = {
```
cesarenaldi, line 344: Question: is dropping `fenced` intentional? The response schema parses it, and it's the field that says whether the key was already blocked off-chain at submission time versus still usable until the transaction confirms. Not saying it must be on the result, I just want to understand if leaving it out was a deliberate call.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+  };
+}
+
+async function waitForAuthorizedSessionKey(
+  client: BaseSecureClient,
+  expected: SessionKey,
+): Promise<SessionKey> {
+  for (
+    let pollCount = 0;
+    pollCount < client.environment.relayerMaxPolls;
```
cesarenaldi, line 449: This loop polls the session-key listing but is bounded by the relayer polling config (`relayerMaxPolls` / `relayerPollFrequencyMs`). The names are misleading at this call site: tuning relayer polling silently changes how long we wait on the listing too.

Either add a neutral polling config for this, or, given this readiness check is temporary, just leave a short code comment stating we deliberately reuse the relayer polling config here.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+    .transform((value): ParsedAuthorizeSessionKeyRequest => {
+      const requestedScopes = new Set(value.scopes);
+
+      return {
+        ...value,
+        address: expectEvmAddress(value.address.toLowerCase()),
+        scopes: [
+          ...SESSION_KEY_KNOWN_SCOPE_ORDER.filter((scope) =>
+            requestedScopes.delete(scope),
+          ),
```
cesarenaldi: We want to surface other scopes that are not modelled in the known enum, this is removing them.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+      }
+    })
+    .transform((value): ParsedAuthorizeSessionKeyRequest => {
+      const requestedScopes = new Set(value.scopes);
+
+      return {
+        ...value,
+        address: expectEvmAddress(value.address.toLowerCase()),
+        scopes: [
+          ...SESSION_KEY_KNOWN_SCOPE_ORDER.filter((scope) =>
```
cesarenaldi: The "known" enum pattern is the one where the field value cannot be exhaustively checked.

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+      if (value.validUntil <= Math.floor(Date.now() / 1_000)) {
+        context.addIssue({
+          code: 'custom',
+          message: 'Session key expiry must be a future Unix timestamp.',
+          path: ['validUntil'],
+        });
+      }
+    })
+    .transform((value): ParsedAuthorizeSessionKeyRequest => {
+      const requestedScopes = new Set(value.scopes);
```
cesarenaldi: This is overly defensive. What is the reason to do this?

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,540 @@
...
+}
+
+function haveSameScopes(
+  left: SessionKeyScope[],
+  right: SessionKeyScope[],
+): boolean {
+  return (
+    left.length === right.length && left.every((scope) => right.includes(scope))
+  );
+}
```
cesarenaldi, line 481: This compares length plus membership, so duplicate entries defeat it: `['CLOB', 'CLOB']` vs `['CLOB', 'ALL']` compares equal. Unreachable today because the request transform dedupes and the server derives scopes from a bitmask, but the helper borrows those invariants from its callers. A set comparison costs the same and holds on its own:

```suggestion
function haveSameScopes(
  left: SessionKeyScope[],
  right: SessionKeyScope[],
): boolean {
  const rightScopes = new Set(right);

  return (
    new Set(left).size === rightScopes.size &&
    left.every((scope) => rightScopes.has(scope))
  );
}
```

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,117 @@
...
+  type EvmAddress,
+  EvmAddressSchema,
+  type TransactionId,
+  TransactionIdSchema,
+  type TxHash,
+  TxHashSchema,
+} from '../shared';
+
+/** Known scope values accepted by session-signer authorization requests. */
+export enum RelayerSessionSignerKnownScope {
```
cesarenaldi: We should not duplicate this enum.

Let's move `SessionSignerKnownScope` and `SessionSignerScope` and `SignerKnownScopeSchema` from client into bindings `shared.ts`.

Drop `RelayerSessionSignerKnownScope`, `RelayerSessionSignerScope`, and `RelayerSessionSignerScopeSchema`.

Use `SessionSignerScope` and `SignerKnownScopeSchema` in place of `RelayerSessionSignerScope` and `RelayerSessionSignerScopeSchema` and in `ActiveSessionSigner.scopes` and its schema which are currently validating just plain strings.

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** Raw operation status returned at submission time. */
+  status: string;
+  /** Identifier used to poll the submitted transaction. */
+  transactionId: TransactionId;
+};
+
+export const RelayerRevokeSessionSignerResponseSchema = z.object({
+  fenced: z.boolean(),
+  operationId: z.string().min(1),
+  status: z.string().min(1),
```
cesarenaldi: this needs an enum I think.

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,117 @@
...
+  operationId: string;
+  /** Raw operation status returned at submission time. */
+  status: string;
+  /** Identifier used to poll the submitted transaction. */
+  transactionId: TransactionId;
+};
+
+export const RelayerRevokeSessionSignerResponseSchema = z.object({
+  fenced: z.boolean(),
+  operationId: z.string().min(1),
```
cesarenaldi: This most likely needs a branded string.

packages/client/src/wallet.ts
```diff
@@ -71,6 +97,62 @@ export function resolveOrderIdentity(account: AccountIdentity): OrderIdentity {
...
+  }
+
+  return account.signer;
+}
+
+/** @internal */
+export function wrapDepositWalletSessionSignerSignature(
+  sessionSigner: EvmAddress,
+  signature: HexString,
+): Erc1271Signature {
```
cesarenaldi: This is not an ERC1271 signature though.

cesarenaldi: Most likely this envelope shape have a name, we can create a new HexString branded type to make the call chain type safe and semantically correct (or this misnomer will be interpreted as correct by future sessions).

packages/client/src/actions/orders/prepare.ts
```diff
@@ -84,7 +85,11 @@ export async function prepareMarketOrder(
 
     return createSignedOrder(
       unsignedOrder,
-      createOrderSignature(unsignedOrder, signature),
+      createOrderSignature(
+        unsignedOrder,
+        signature,
+        resolveDepositWalletSessionSigner(client.environment, client.account),
```
cesarenaldi: This is is relatively computationally expensive to do it for every single order.

We should define a `SignarType` enum:

```
enum SignerType {
  OWNER
  SESSION_KEY
}
```

and expand `AccountIdentity` to have a `signerType: SignerType` property.

packages/client/src/actions/orders/prepare.ts
```diff
@@ -84,7 +85,11 @@ export async function prepareMarketOrder(
 
     return createSignedOrder(
       unsignedOrder,
-      createOrderSignature(unsignedOrder, signature),
+      createOrderSignature(
```
cesarenaldi: We should reuse `wrapDepositWalletSignature` here rather than have `createOrderSignature` do the same job as `wrapDepositWalletSignature`.

packages/client/src/wallet.ts
cesarenaldi: I asked protocol team. They do n't have a formalized name and due the use of the ERC6492_MAGIC we refer to it as "6492" although it's a misnomer.

We need to come up with a meaningful name that then we use consistently in docs. What about simple and neat: `SessionSignerSignature`?

packages/client/src/actions/orders/prepare.ts
cesarenaldi:
```
wrapDepositWalletSignature(client, createOrderSignature(unsignedOrder, signature))
```

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,542 @@
...
+
+function createAuthorizeSessionKeyRequestSchema(wallet: EvmAddress) {
+  const schema = z
+    .object({
+      address: EvmAddressSchema,
+      idempotencyKey: z.string().trim().min(1).optional(),
+      scopes: z
+        .array(SessionKeyScopeSchema)
+        .min(1)
+        .optional()
```
cesarenaldi:
```suggestion
        .default(DEFAULT_SESSION_KEY_SCOPES)
```

packages/client/src/actions/session-keys.ts
```diff
@@ -0,0 +1,542 @@
...
+const DEFAULT_SESSION_KEY_SCOPES = [SessionKeyKnownScope.ALL] as const;
+
+type ParsedAuthorizeSessionKeyRequest = {
+  address: EvmAddress;
+  idempotencyKey?: string;
+  scopes: SessionKeyScope[];
+  validUntil: number;
+};
+
+function createAuthorizeSessionKeyRequestSchema(wallet: EvmAddress) {
```
cesarenaldi: We don't need a factory function, it can be `const AuthorizeSessionKeyRequestSchema = z.... satisfies ... `

packages/client/src/actions/gasless.ts
```diff
@@ -633,21 +644,38 @@ export async function* resignDepositWalletExecuteRequest(
...
+          wallet: client.account.wallet,
+        }),
+      ),
     ),
   );
 
   return { ...request, nonce, signature };
 }
 
+function wrapDepositWalletSignature(
```
cesarenaldi: Move this to wallet.ts

packages/client/src/actions/gasless.ts
```diff
@@ -633,21 +644,38 @@ export async function* resignDepositWalletExecuteRequest(
...
+  signature: EvmSignature,
+) {
+  const sessionSigner = resolveDepositWalletSessionSigner(
+    client.environment,
+    client.account,
+  );
+
+  return sessionSigner === undefined
+    ? signature
+    : wrapDepositWalletSessionSignerSignature(sessionSigner, signature);
```
cesarenaldi: Once `wrapDepositWalletSignature` direct usage of `wrapDepositWalletSessionSignerSignature` should not longer be needed and `wrapDepositWalletSessionSignerSignature` become private to the wallet.ts module.

packages/client/src/wallet.ts
```diff
@@ -23,17 +25,41 @@ export type AccountIdentity = {
...
+  '0x0000000000000000000000000000000000000000000000000000000000000000' satisfies HexString;
+const SESSION_SIGNER_MAGIC_BYTES =
+  '0x6492649264926492649264926492649264926492649264926492649264926492' satisfies HexString;
+
 /** @internal */
 export function resolveAccountIdentity(
   environment: EnvironmentConfig,
   signer: EvmAddress,
   wallet: EvmAddress,
 ): AccountIdentity {
```
cesarenaldi, line 49:
```suggestion
export async function resolveAccountIdentity(
  client: BaseClient,
  signer: EvmAddress,
  wallet: EvmAddress,
): Promise<AccountIdentity>
```

packages/client/src/wallet.ts
```diff
@@ -2,38 +2,63 @@ import { SignatureType } from '@polymarket/bindings/clob';
...
-  return {
-    signer,
-    wallet,
-    walletType: classifyWalletType(environment, signer, wallet),
-  };
+  return tryResolveAccountIdentity(environment, signer, wallet);
+}
+
+/** @internal */
+export function tryResolveAccountIdentity(
```
cesarenaldi: There is no longer reason to have `tryResolveAccountIdentity`, let's move code into `resolveAccountIdentity`

packages/client/src/actions/orders/types.ts
```diff
@@ -193,7 +199,7 @@ export type SignedOrder = {
    * named `tokenId` for both CTF token IDs and Polymarket V2 position IDs.
    */
   tokenId: PositionId | TokenId;
-  signature: EvmSignature | Erc1271Signature;
+  signature: OrderSignature;
```
cesarenaldi, line 202: 👍

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+import {
+  type EvmAddress,
+  type SessionSignerScope,
+  type TransactionId,
+  TransactionIdSchema,
+  type TxHash,
+  TxHashSchema,
+} from '../shared';
+
+export type RelayerAuthorizeSessionSignerRequest = {
```
cesarenaldi: Not used.

packages/bindings/src/relayer/session-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// The upstream `operationId` is omitted from the normalized response because
+// it is an opaque identifier with no SDK consumer.
+export const RelayerAuthorizeSessionSignerResponseSchema = z.object({
+  status: z.enum(RelayerAuthorizeSessionSignerStatus),
+  transactionHash: TxHashSchema.nullish().transform((value) => value ?? null),
+  transactionId: TransactionIdSchema,
+}) satisfies z.ZodType<RelayerAuthorizeSessionSignerResponse>;
+
+export type RelayerRevokeSessionSignerRequest = {
```
cesarenaldi: Drop this, no request shape definitions needed in bindings.

cesarenaldi (approved):

packages/client/src/actions/session-keys.ts
kartojal: yes it is intentional, added a comment about it

cesarenaldi (approved):

PR #313, opened 2026-08-26
feat(data-api): sdk core client envelope errors 429 retry auth ready

ExoMonk:
## feat(data): v2 envelope schemas, rate-limit retry helper, MIGRATION activity type

Core pieces so the client can adopt the data `/v2` surface (public since 2026-08-25 at `data-api.polymarket.com/v2`). **Everything is additive — no existing API shape changes.** The transport, error taxonomy, auth header slot, environments, and the `paginate()` walker already fit `/v2` as-is; this PR adds only what was genuinely missing.

Closes DEV-609.

### v2 envelope schemas (`@polymarket/bindings/data`)

- `dataV2PageSchema(item)` — the paginated list envelope `{ data[], pagination { limit, offset, has_more, next_cursor } }`, normalized to `hasMore` / branded `PaginationCursor` (`null` → `undefined`).
- `dataV2EnvelopeSchema(payload)` — the single-object envelope `{ data }`; endpoints whose ordinary answer is "no such row" serve `data: null`, modeled by passing a nullable payload schema so absence stays a parsed value.
- The schema **enforces the pagination invariant** `has_more === (next_cursor !== null)` (and rejects an empty-string cursor). The invariant's silent failure mode is nasty: `has_more: true` with no cursor hands the page walker `undefined`, restarting it from page 1 forever. A broken upstream invariant now fails validation loudly instead.

### `withRateLimitRetry` (`@polymarket/client`, internal)

Rate limiting is ordinary backpressure on this surface (the heavy-query gate and pool saturation both answer 429 + `Retry-After`), so idempotent reads should absorb it rather than surface it on the first hit.

- Retries only `RateLimitError`; every other error propagates immediately.
- **The wait is exactly what the server asked, or nothing**: `retryAfter` honored as-is (1s assumed when absent); a requested delay longer than `maxDelaySeconds` (default 5) propagates rather than retrying early into a window the server declared closed. Defaults: 2 retries.
- Injectable time seam; not exported from the package root — actions compose it (reads only) starting with the v2 endpoint work.

### `ActivityType.MIGRATION`

The v2 activity taxonomy includes `MIGRATION` (protocol-v2 position migrations). Adding it turned out to be load-bearing: the exhaustive `normalizeActivity` switch failed to compile until a proper `MigrationActivity` variant joined the `Activity` union. It sits in the plain-amount branch deliberately — migration rows can carry an empty condition id (event-fallback titles), so the market-linked branch's `expectPresent(conditionId)` would reject real rows. Verified against the live v1 endpoint: `?type=MIGRATION` answers 200, so the new member cannot break existing v1 calls.

### Tests

- `packages/bindings/src/data/v2.test.ts` — pagination normalization + all three inconsistent-envelope shapes rejected (5 tests).
- `packages/client/src/retry.test.ts` — honored delay, 1s default, over-cap propagation, exhaustion, non-rate-limit passthrough (5 tests).
- `packages/client/tests/integration/data-v2.test.ts` — **live against the public `/v2`**: a paginated envelope with a real server-minted cursor (leaderboard), a single-object envelope (oi), and the `{ data: null }` unknown-wallet answer (user-stats). 3/3.

### Verification

- `pnpm build` — clean
- `pnpm lint` / `pnpm typecheck` — clean
- `pnpm test:bindings` — 119 passed
- `pnpm test:client` — 386 passed
- `pnpm test:client:integration data-v2` — 3 passed (live network)
- Changeset included: minor for `@polymarket/bindings` and `@polymarket/client` (the enum member and its variant reach the client's public re-exports).

---

> [!NOTE]
> **Low Risk**
> Changes are additive (schemas, internal retry helper, new activity enum variant) with no modifications to existing API contracts; integration tests hit live `/v2` read-only endpoints.
> 
> **Overview**
> Adds foundational SDK pieces for the data API **`/v2`** surface without changing existing v1 shapes.
> 
> **Bindings** exports `dataV2PageSchema` and `dataV2EnvelopeSchema`, mapping wire `{ data, pagination }` into the SDK **`Page`** shape (`items`, `hasMore`, branded `nextCursor`) and unwrapping single-object responses (including **`data: null`** when the payload schema is nullable). Pagination parsing **rejects inconsistent `has_more` / `next_cursor`** so a broken upstream response cannot silently restart cursor pagination.
> 
> **Client** adds internal **`withRateLimitRetry`**, which retries only **`RateLimitError`**, waits for server **`retryAfter`** (default 1s, cap 5s before failing through), and leaves other errors untouched—intended for idempotent reads on the v2 surface.
> 
> **Activity** gains **`ActivityType.MIGRATION`** and **`MigrationActivity`** (USD `amount`), wired through **`normalizeActivity`** in the plain-amount branch so rows with empty condition ids still parse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d02b3374f550f5eb00ada4c7f7e5177ab0fb4f51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/retry.ts
```diff
@@ -0,0 +1,69 @@
...
+   * limit asking for a longer delay propagates instead of being retried —
+   * retrying earlier than the server asked would burn the attempt into a
+   * window the server already declared closed. Defaults to 5.
+   */
+  maxDelaySeconds?: number;
+  /** @internal Test seam for controlled time. */
+  sleep?: SleepFn;
+};
+
+function defaultSleep(milliseconds: number): Promise<void> {
```
cesarenaldi: nit: there is a `delay` helper in:
```
import { delay } from "@polymarket/types";
```

packages/bindings/src/data/v2.ts
cesarenaldi, line 1: We normalize all paginated endpoint under the same pagination interface to hide the seams across the APIs.

The pagination envelope is already defined as `Page<T>`.

No matter what is the underlying pagination strategy the SDK always exposes an iterable with cursor-like pagination.

ExoMonk: Thanks, I will make sure to use this type

packages/client/tests/integration/data-v2.test.ts
cesarenaldi, line 1: I see this file has been used to test the intermediate helper which is fine for now.

Integration tests in this repo are intended to test the same surface external integrators will use. I expect tests in this file to evolve. Actually, as we replace v1 endpoints with v2, you will find plenty of existing integration tests against Data API that can be adapted to v2.

ExoMonk: Agreed — the helpers are the test subject here only because this slice ships no v2 action yet, so there's no public surface to integrate against. The next slice (first real v2 endpoint wired through an action) flips that: integration coverage moves to the client surface (client.data.list* + paginate()), these envelope-level tests fold into it, and we'll adapt the existing v1 Data API integration tests endpoint-by-endpoint as each one moves to v2.

cesarenaldi (approved):

PR #315, opened 2026-08-26
feat(data-api-v2): Cursor implementation with /v2/trades replacement - DEV-610

ExoMonk:
## What

DEV-610 — the cursor pagination engine, proven end-to-end by fully replacing `listTrades` with the new Data API contract. Same method name, new behavior — per the team decision: the SDK is pre-release, so endpoints land as in-place replacements with no "V2" suffix anywhere.

## Breaking changes

- `listTrades` now serves exact continuation signals: `hasMore` is authoritative and `nextCursor` is server-minted, so the end of a collection never costs an extra probe request. `pageSize` defaults to 100 (max 1000 — larger values are rejected, not clamped; previously default 20, max 10000). Offset does not exist in the vocabulary.
- `Trade` is now the strict normalized row: `size`/`price` are numbers, `timestamp` is epoch milliseconds, and absence is uniform — empty-string text fields and the unknown-outcome sentinel (`outcome_index: 999`) both parse to `undefined`. Field names are unchanged (`wallet`, `conditionId`, `tokenId`, …), which the pre-existing `listTrades` integration tests prove: they pass unchanged against the replacement.
- `filterType`/`filterAmount` may now be sent alone (the service always applies the dust filter and fills the missing half — TOKENS / 0.01); the old both-or-neither client error is gone. Empty `user`/`market`/`eventId` filters are rejected loudly instead of forwarded: the service reads an empty `user` as absent and keys routing on `market`/`event_id` presence, so forwarding them silently widens the request to the global feed.
- Transient rate limits are absorbed on the read path by retrying after the server-requested delay (first real use of `withRateLimitRetry`).

## Why this shape

The trades cursor binds only its paging anchor (limit, timestamp, sequence_id, direction) — filters are NOT in it. The action's fetchPage closure therefore re-sends the complete original param set with every cursor, which makes two failure modes unreachable by construction: silently widened cohorts from a filter dropped on page 2, and the pinned-cursor conflict. `limit` riding along with a cursor is explicitly ignored by the service (the cursor's size wins), so the unconditional resend is safe.

## Tests

Integration tests exercise the surface external integrators use: a 3-page `for await` walk proving filter re-send (a side filter holds on every page — the cursor doesn't carry it), descending keyset order across the whole walk, distinct server-minted cursors, and a `from(cursor)` continuation (`tests/integration/data-pagination.test.ts`). The pre-existing `listTrades` integration tests run unchanged against the replacement. Envelope-level tests remain only for surfaces with no public action yet and fold into their action's test as each lands.

---

> [!NOTE]
> **Medium Risk**
> Breaking public API for `listTrades` and `Trade` types affects integrators; pagination and filter semantics changes can alter result sets if callers relied on `market`, offset behavior, or loose optional fields.
> 
> **Overview**
> **Breaking:** `listTrades` keeps the same name but targets **`/v2/trades`** with a new contract. Pagination is **cursor-based** (`hasMore` + server-minted `nextCursor`) instead of client offset cursors and page-length probing; **every page re-sends the full filter set** because the cursor only anchors position. **`pageSize`** defaults to **100** (max **1000**, rejected if higher). Request filters use **`conditionId`** instead of **`market`**, with stricter validation (branded `user`, deduped condition ids capped at 20, `conditionId` xor `eventId`). **`filterType` / `filterAmount`** may be sent independently. Transient **rate limits** are retried via `withRateLimitRetry`.
> 
> Bindings add reusable **`dataPageSchema`** / **`dataEnvelopeSchema`** (renamed from the old `v2` helpers); **`ListTradesResponseSchema`** is now a paginated envelope, not a bare array. **`Trade`** is parsed from strict snake_case wire rows into a tighter SDK shape: required core fields, **`size`/`price` as numbers**, **`timestamp` in ms**, empty strings and **`outcome_index: 999`** → **`undefined`**.
> 
> Legacy data actions switch query encoding to **`toLegacyDataSearchParams`**; **`toDataSearchParams`** snake_cases keys for the new v2-style calls. Integration coverage moves to **`data-pagination.test.ts`** (multi-page `listTrades` + envelope smoke tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2195b8f7b5f13a624d116a968b78195b102ea9b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/actions/activity-v2.ts
```diff
@@ -0,0 +1,140 @@
...
+  })
+  .refine((value) => !(value.market && value.eventId), {
+    message: 'Provide market or eventId, not both',
+    path: ['eventId'],
+  })
+  // A zero `end` means unbounded, mirroring the service.
+  .refine((value) => !value.end || value.end >= (value.start ?? 0), {
+    message: 'end must not precede start',
+    path: ['end'],
+  });
```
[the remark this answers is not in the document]
ExoMonk: Deliberate, and verified live rather than assumed: unlike v1, the v2 service legalized partial filters — it always applies the dust filter and fills the missing half in (TOKENS / 0.01). filter_type alone and filter_amount alone both answer 200 on prod today. Re-adding v1's both-or-neither rule would make the SDK reject requests the service answers. The schema now carries a comment stating this (with the verification), and the TSDoc says "either may be sent alone."

packages/client/src/actions/activity.ts
```diff
@@ -40,30 +46,37 @@ const TradeFilterTypeSchema = z.enum(['CASH', 'TOKENS']);
...
     filterType: TradeFilterTypeSchema.optional(),
-    filterAmount: z.number().optional(),
-    market: z.array(z.string()).optional(),
-    eventId: z.array(z.number().int()).optional(),
-    user: z.string().optional(),
+    filterAmount: z.number().min(0).optional(),
+    // Encodes to `condition_id`, an accepted alias of the canonical
+    // `condition` key. At most 20 distinct ids — the service caps the
+    // selector (DENG-588), and enforcing it here keeps the error typed.
+    conditionId: z.array(z.string().min(1)).min(1).max(20).optional(),
```
cesarenaldi:
```suggestion
    conditionId: z.array(ConditionIdSchema).min(1).max(20).optional(),
```

packages/client/src/actions/activity.ts
```diff
@@ -40,30 +46,37 @@ const TradeFilterTypeSchema = z.enum(['CASH', 'TOKENS']);
...
+    eventId: z.array(z.number().int()).min(1).optional(),
```
cesarenaldi: Existing code already suffer of the following issue, just noticing this only now.

```suggestion
    eventId: z.array(EventIdSchema).min(1).optional(),
```

Context: in general we do have branded types for many literal types that help with type safety (`EvmAddress`, `EventId`, `MarketId`, etc.) for each one we have a corresponding zod schema that asserts their type. When used in request objects like here they validate user's input and help with developer friendly error messages that pin point any mistake.

In the specific of `EventIdSchema` I looked into it and remembered that we built-it so that accommodates for another discrepancy between Data API v1 schema and other API surfaces.

```
const EventIdSchema = z
  .union([z.string(), z.number().int().transform(String)])
  .transform(toEventId);
```

It basically normalizes Event ID to be strings, this is based on the fact that Gamma, the authoritative source for Event ID, encodes its value as as string (same that it does for Market ID for example).

In Data API event Ids are instead JS numbers on the wire. Is this a fix you would consider for the Data API v2?

packages/client/src/actions/params.ts
```diff
@@ -81,6 +83,35 @@ export function toDataSearchParams<TParams extends SearchParamsInput>(
   return searchParams;
 }
 
+/**
+ * Data endpoints use snake_case query keys and comma-separated arrays.
+ */
+export function toDataSearchParams<TParams extends SearchParamsInput>(
```
cesarenaldi, line 89: 👍

packages/client/src/decorators/data.ts
```diff
@@ -275,7 +275,19 @@ export type DataActions = {
...
+   * Pagination is cursor-only: `hasMore` is exact and `nextCursor` is
+   * server-minted, so the end of the collection never costs an extra request.
+   * Every page re-sends the original filters. `pageSize` defaults to 100 (at
+   * most 1000 — larger values are rejected, not clamped). Only the taker side
+   * of each match is returned by default (`takerOnly: false` includes maker
+   * rows), and a dust filter of 0.01 shares applies unless
+   * `filterType`/`filterAmount` say otherwise (either may be sent alone —
+   * the service fills the other half in). `conditionId` accepts at most 20
+   * ids. `start`/`end` are Unix seconds. Transient rate limits are absorbed
+   * by retrying after the server-requested delay.
```
cesarenaldi: Too much technical details here. SDK consumer are not exposed to the details and discrepancy of pagination approaches across APIs.

I would omit anything regarding pagination.

packages/client/src/actions/activity.ts
```diff
@@ -127,33 +156,18 @@ export function listTrades(
...
-                offset: decoded.offset + decoded.pageSize,
-                pageSize: decoded.pageSize,
-              })
-            : undefined,
-        };
-      });
-  }, cursor);
+      ).andThen(validateWith(ListTradesResponseSchema)),
+    cursor,
+  );
```
cesarenaldi, line 176: Much better with native cursor pagination support!

cesarenaldi (approved):
PR looks pretty good to me.

I left few DX improvement comments, some of which stems from the previous code already having an issue.

Happy to pick up this separately but maybe by instructing your agent we can fix those small issues as we go.

Pre-approved, feel free to merge once ready.

PR #325, opened 2026-08-27
feat: integrate positions & activity with combos for data-api /v2 path

ExoMonk:
## What

DEV-611 — the high-traffic feeds, fully replaced in place on the DEV-610 template: listActivity, listPositions, listComboActivity, listComboPositions. Same names, new contracts; net −485 lines.

## Breaking changes

- listPositions bundles the whole lifecycle behind status (OPEN default ⊇ REDEEMABLE, CLOSED) with redeemable/mergeable flags and fee-exclusive entry economics on every row. listClosedPositions and listMarketPositions are removed — use status: 'CLOSED' and the single-conditionId anchor arm respectively.
- Every windowed method (trades included, retrofitted here) takes one window option — 'full' | { start?, end? }, accepting epoch seconds or Date — replacing raw start/end.
- The Activity union, Position, ComboPosition, and ComboActivity rows are strict and normalized: numbers for amounts, epoch milliseconds, empty-string/999-sentinel absence as undefined, assetId canonical with a deprecated tokenId alias. Combo activity rows carry positionId on every row and drop transactionAt/logIndex/moduleId with the wire; exact combo economics (grossEntryCostUsdc, entryFeesUsdc) stay decimal strings by design.
- listComboPositions no longer accepts updatedAfter/updatedBefore — the service rejects them as not yet ported (DENG-589 tracks porting the watermark; the SDK re-adds both in one line after).
- Condition filters share one dedupe-then-cap rule (distinctIdList): 20 distinct ids on trades (DENG-588's service cap), 100 elsewhere (headroom under the measured ~8KB URL edge), always as typed UserInputErrors.

## Not changed

fetchPortfolioValue, fetchTradedMarketCount, downloadAccountingSnapshot, listOpenInterest, listMarketHolders stay on their current endpoints (DEV-612's slice).

## Tests

Existing integration tests adapted rather than duplicated (per review guidance on #313): both positions status arms live; combo status-CSV serialization and cursor resend proven with fetch spies against /v2/positions/combos; the activity union parses live prod rows; the deposits/withdrawals opt-out is unit-tested on the v2 path with the page envelope.

---

> [!NOTE]
> **High Risk**
> Major breaking SDK and wire contract changes across portfolio, activity, and pagination paths that will break existing integrations until callers migrate to v2 shapes, filters, and removed methods.
> 
> **Overview**
> **Breaking** refresh of the high-traffic data feeds: `listActivity`, `listComboActivity`, `listPositions`, and `listComboPositions` keep the same names but call **`/v2/*`**, use **cursor pagination** (default page size **100**), **`withRateLimitRetry`**, and shared **`toDataSearchParams`** / **`dataPageSchema`** envelopes instead of offset paging or bare arrays.
> 
> **Positions:** `listPositions` is the single lifecycle API — filter with **`status`** (`OPEN` / `REDEEMABLE` / `CLOSED`), **`conditionId`** (replaces `market`), dust/sort/window options, and stricter rows (`currentSize`, `currentPrice`, fee-split **`entryCostUsdc`** / **`entryFeesUsdc`**, PnL fields as **`DecimalString`**). **`listClosedPositions`** and **`listMarketPositions`** are removed (use `status: 'CLOSED'` or a lone **`conditionId`** without `user` on a public client). Authenticated **`listPositions`** always injects the account wallet.
> 
> **Activity & combos:** Wire parsing moves to strict **snake_case**; trades use **`OrderSide`** ( **`Side`** removed); **`TIP`** / **`TipSide`** added. Combo activity rows always expose **`positionId`** and drop **`transactionAt`**, **`logIndex`**, **`moduleId`**. Combo positions swap YES/NO **`side`** for **`outcomeIndex`** / **`outcomeLabel`**, add **`Redeemable`** status, and use **`sortBy`** / **`sortDirection`** (replacing **`ComboPositionSort`**).
> 
> **Cross-cutting:** History bounds use one **`window`** (`'full'` or `{ start?, end? }` with seconds/`Date`) instead of raw **`start`**/**`end`**. **`conditionId`** lists are validated with **`distinctIdList`** (≤20 distinct ids, typed **`UserInputError`**). Bindings drop **`ClosedPosition`** / **`MarketPosition`** types and normalize amounts/timestamps consistently.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d88606866c6abbc2bb8e096c264277e500594ea1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/decorators/account.ts
```diff
@@ -57,10 +51,17 @@ type DefaultAccountWallet<TRequest extends { user: string }> = Prettify<
...
-  DefaultAccountWallet<ListClosedPositionsRequest>;
+export type SecureListPositionsRequest = Prettify<
+  Omit<ListPositionsRequest, 'user'> & {
+    /**
+     * Wallet address to use. Pass `null` to anchor on a market's holders
+     * (single `conditionId`, no wallet) instead of a wallet's positions.
+     *
+     * @defaultValue `client.account.wallet`
+     */
+    user?: string | null;
```
cesarenaldi: `user` now has three meanings: omitted defaults to the authenticated wallet, a string selects a wallet, and `null` suppresses wallet binding to switch into the market-holders workflow. This is inconsistent with the other `SecureClient` bindings, difficult to discover from the call site, and error-prone in JavaScript, where `null` and `undefined` are often used interchangeably. Like those other methods, this should default an omitted `user` to the authenticated account and accept an explicit address as an override. Consumers who need positions for a market without a user filter can use `PublicClient.listPositions({ conditionId })`.

packages/bindings/src/data/common.ts
```diff
@@ -14,12 +14,17 @@ export enum ActivityType {
...
+  /** A user-to-user tip, served in the feed like every other type. */
+  TIP = 'TIP',
 }
 
 export const ActivityTypeSchema = z.enum(ActivityType);
 
 export const SideSchema = z.enum(['BUY', 'SELL']);
 
+/** Direction of a TIP from the row wallet's perspective. */
+export const TipSideSchema = z.enum(['IN', 'OUT']);
```
cesarenaldi: The TypeScript SDK represents vocabularies as exported string enums so consumers can discover supported values through autocomplete, while `z.enum(Enum)` keeps runtime validation and public types synchronized. Can we apply that consistently here?

For `BUY | SELL`, please reuse the existing `OrderSide` and `OrderSideSchema` across `listTrades`, `listActivity`, and `TradeActivity.side` rather than introducing a duplicate `Side` type.

The new vocabularies should likewise be exported enums: `TipSide`, `SortDirection`, `TradeFilterType`, `PositionFilterType`, `PositionSortBy`, and `ComboPositionSortBy`.

The intended consumer DX would be:

```ts
client.listTrades({
  side: OrderSide.BUY,
});

client.listPositions({
  status: PositionStatus.Closed,
  filterType: PositionFilterType.Cash,
  sortBy: PositionSortBy.Timestamp,
  sortDirection: SortDirection.Desc,
});

if (
  activity.type === ActivityType.TIP &&
  activity.side === TipSide.In
) {
  // ...
}
```

With schemas derived from the same definitions:

```ts
export const TipSideSchema = z.enum(TipSide);
export const SortDirectionSchema = z.enum(SortDirection);
```

packages/bindings/src/data/portfolio.ts
```diff
@@ -32,213 +26,187 @@ export enum ComboPositionStatus {
...
-    outcome: z.string().nullish(),
-    outcomeIndex: z.number().int().nullish(),
-    oppositeOutcome: z.string().nullish(),
-    oppositeAsset: ClobAssetIdSchema.nullish(),
-    endDate: IsoCalendarDateStringSchema.nullish(),
-    negativeRisk: z.boolean().nullish(),
+    proxy_wallet: EvmAddressSchema,
+    token_id: ClobAssetIdSchema,
+    condition_id: ConditionIdSchema,
+    current_size: z.number(),
```
cesarenaldi: I understand Data API v2 changes many fractional values that were previously decimal strings into JSON numbers, for example the position economics in `PositionSchema` and `size` and `price` in the activity bindings.

The goal of unification is to expose consistent platform types across API seams. For prices, amounts, sizes, fees, and PnL, that shared type should remain `DecimalString`.

A correct decimal can lose precision immediately when deserialized:

```ts
JSON.parse('{"amount":0.123456789012345678}').amount;
// 0.12345678901234568
```

Once that happens, the original value cannot be recovered. A decimal string preserves it across language boundaries, while consumers can still explicitly convert it to a number when acceptable.

Can we keep exact fractional values as decimal strings in the v2 backend contract?

cesarenaldi (changes requested):

packages/client/src/decorators/account.ts
ExoMonk: Agreed and done. SecureListPositionsRequest is back on the DefaultAccountWallet pattern (omitted user = authenticated wallet, string = override); the null arm and withOptionalAccountWallet are removed. Market-holders anchoring lives on the public client's listPositions, and the secure method's TSDoc says so.

packages/bindings/src/data/common.ts
ExoMonk: Done. Side/SideSchema are removed in favor of the existing OrderSide/OrderSideSchema (listTrades, listActivity, TradeActivity.side), and TipSide, SortDirection, TradeFilterType, PositionFilterType, PositionSortBy, and ComboPositionSortBy are now exported enums with z.enum(Enum), re-exported through the client the same way as PositionStatus/ComboPositionStatus. Your DX example compiles as written.

packages/bindings/src/data/portfolio.ts
ExoMonk: Good catch, and it sent us to verify the wire. Turns out v1 itself serves JSON numbers ("size":982568.9716, "value":246881.3079 on prod right now); the DecimalString the old SDK returned was client-side coercion of those numbers, never a wire fact.

Decision (DENG-599): JSON numbers everywhere, true v1 wire parity. The two combo fields that serve strings today (gross_entry_cost_usdc, entry_fees_usdc) flip to numbers so /v2 has one encoding.

Why numbers lose nothing here: every money/size value on this surface is 6-dp grain (micro-USDC), served from NUMERIC columns with a single cast to double at projection. Distinct 6-dp values map to distinct doubles up to ~4.5e9 (1e-6 grain vs 2^-52 relative spacing), and platform magnitudes top out around 1e8, so formatting the double back to 6 dp recovers the stored value exactly. Your 0.123456789012345678 example is an 18-dp value; nothing on this surface carries that grain, and a value that lost precision at the service's own f64 serialization could not be recovered by wrapping it in a string. Integers stay integers, booleans stay booleans, percents stay numbers.

The single-cast invariant, the 6-dp grain per field, and round-trip spec tests at boundary magnitudes ride DENG-599 so this stays true by construction. So the SDK keeps number for these fields with the grain documented on the types; consumers doing decimal arithmetic across thousands of rows should use a decimal lib client-side regardless of wire encoding, and the TSDoc will say so.

cesarenaldi: I understand we go live with JS Number like in v1. In that case let's re-use `DecimalishSchema` (and `DecimalString` on the type) for all these fields so that the DX on the SDK surface feels even also on v2.

packages/bindings/src/data/portfolio.ts
```diff
@@ -32,213 +26,217 @@ export enum ComboPositionStatus {
...
+  oppositeOutcome?: string;
   /** Opposite outcome identifier for a CTF token or Polymarket V2 position. */
-  oppositeAssetId: TokenId | PositionId | null | undefined;
+  oppositeAssetId?: TokenId | PositionId;
   /** @deprecated Use `oppositeAssetId`. */
-  oppositeTokenId: TokenId | PositionId | null | undefined;
-  endDate?: IsoCalendarDateString | null;
-  negativeRisk?: boolean | null;
+  oppositeTokenId?: TokenId | PositionId;
+  endDate?: string;
```
cesarenaldi:
```suggestion
  endDate?: IsoCalendarDateString;
```

and corresponding field schema.

packages/bindings/src/data/portfolio.ts
ExoMonk: Done in 2f648a9 — every money/size/price/PnL value on the v2 surface now normalizes through DecimalishSchema to DecimalString, so v2 reads like the v1 SDK surface. The wire's JSON numbers stringify losslessly at their 6-decimal grain, and Decimalish tolerates both encodings, which also makes the SDK transition-proof across the service-side combos encoding flip that is deploying now (DENG-599).

packages/bindings/src/data/portfolio.ts
ExoMonk: Done in 2f648a9 — endDate is IsoCalendarDateString with the corresponding field schema (empty-string absence still normalizes to undefined).

packages/client/src/actions/portfolio.ts
```diff
@@ -143,190 +179,89 @@ export function listPositions(
...
   | readonly [ComboPositionStatus, ...ComboPositionStatus[]];
 
 const ComboPositionStatusFilterSchema = z.union([
   ComboPositionStatusSchema,
-  z.tuple([ComboPositionStatusSchema], ComboPositionStatusSchema),
+  z
+    .tuple([ComboPositionStatusSchema], ComboPositionStatusSchema)
+    .refine((statuses) => !statuses.includes(ComboPositionStatus.Redeemable), {
+      message: 'REDEEMABLE must be the sole status value',
+    }),
```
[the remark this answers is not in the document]
ExoMonk: Right — fixed in d886068: the refine now rejects REDEEMABLE only when it rides with other statuses, so the type-valid sole-element array passes and serializes identically to the scalar form.

PR #326, opened 2026-08-27
fix(client): default session key expiration

kartojal:
Removes validUntil parameter in session keys, now Session Keys have a fixed lifetime of 180 days.

---

> [!NOTE]
> **Medium Risk**
> Removing `validUntil` from the public request is a breaking change for SDK consumers, and fixed 180-day lifetimes affect how integrators manage session-key rotation and least-privilege expiry.
> 
> **Overview**
> **Breaking API change:** `authorizeSessionKey` no longer accepts `validUntil`. Callers only pass `address` (plus optional `idempotencyKey` / `scopes`).
> 
> The client now sets expiry internally to **180 days** from authorization time via `SESSION_KEY_LIFETIME_SECONDS`, uses that value for the on-chain call and relayer payload, and drops the “must be a future Unix timestamp” validation. Docs and examples are updated accordingly; integration tests assert the returned `sessionKey.validUntil` falls within that window instead of passing a custom expiry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f21a85c1866c1dffe80f1e3cafc3aac8aeba496. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/actions/session-keys.ts
```diff
@@ -121,20 +121,18 @@ export type AuthorizeSessionKeyRequest = {
...
   address: EvmAddress;
   idempotencyKey?: string;
   scopes: SessionKeyScope[];
-  validUntil: number;
 };
 
 const DEFAULT_SESSION_KEY_SCOPES = [
   SessionKeyKnownScope.ALL,
 ] satisfies SessionKeyScope[];
+const SESSION_KEY_LIFETIME_SECONDS = 180 * 24 * 60 * 60;
```
cesarenaldi: This uses the exact 4,320-hour relayer ceiling, so a client clock ahead of the relayer can push the request over the maximum. Can we keep the fixed 180-day product contract but derive `validUntil` with the original 4,315-hour value to preserve the clock-skew buffer?

cesarenaldi (approved):

PR #330, opened 2026-08-28
feat(client): support Perps internal transfers

kartojal:
## Summary

- add owner-signed `SecureClient.transferPerpsCollateral` with one-attempt unknown-outcome semantics
- add normalized `PerpsSession.listInternalTransfers` pagination and forward-compatible transfer types
- preserve machine-readable non-400 rejection identifiers and add release changesets

## Contract notes

- internal-transfer amounts are decimalized token units; `"100.00"` is signed and sent without base-unit conversion
- repeated signatures are replay-rejected, not idempotently replayed, so the action exposes no retry control
- history pagination deduplicates inclusive timestamp boundaries; the upstream timestamp-only 500-row limit remains documented in DEV-628

## Verification

- `pnpm build`
- `pnpm test` (528 tests passed; client type tests clean)
- `pnpm lint`
- `pnpm typecheck`

Linear: [DEV-628](https://linear.app/polymarket/issue/DEV-628/support-owner-signed-perps-internal-transfers-and-history-in-ts-sdk)
Upstream: [perpetuals#1270](https://github.com/Polymarket/perpetuals/pull/1270)

---

> [!NOTE]
> **Medium Risk**
> Introduces owner-signed collateral movement between accounts; mistakes or duplicate submits could move funds, though owner-only signing, input validation, and single-attempt semantics limit exposure.
> 
> **Overview**
> Adds **owner-signed Perps collateral transfers** on the secure client via `transferPerpsCollateral`, wired only for owner signers (not session keys). The flow signs an `internalTransfer` op with the **decimal amount unchanged**, posts once to `/v1/account/internal-transfer`, and **does not retry** on failure so ambiguous outcomes are reconciled via optional labels and history.
> 
> Adds **`PerpsSession.listInternalTransfers`** with the same descending history pagination/deduping as deposits and withdrawals. Bindings now parse snake_case API payloads into normalized records, including forward-compatible **`type`** values (`transfer`, `referral_payout`, or unknown strings).
> 
> **`ServiceClient`** now treats snake_case JSON `error` strings on non-400 responses as `RequestRejectedError.code` when no explicit `code` is present (e.g. `signer_does_not_match_account`), while keeping explicit codes authoritative.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a582779e5e476d736abfeb947c3e6c0c0cb27fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fresh-perps-transfers.md
+++ b/.changeset/fresh-perps-transfers.md
@@ -1,0 +1,6 @@
+---
+'@polymarket/bindings': minor
+'@polymarket/client': minor
+---
+
+Add owner-signed Perps internal transfers and normalized transfer history.
--- a/packages/bindings/src/perps/common.ts
+++ b/packages/bindings/src/perps/common.ts
@@ -239,6 +239,32 @@ export enum PerpsKnownWithdrawalStatus {
 export type PerpsWithdrawalStatus = PerpsKnownWithdrawalStatus | (string & {});
 
 /**
+ * Known internal-transfer classifications.
+ *
+ * The service evolves this set independently of released clients, so
+ * transfer parsing accepts unknown classifications as plain strings; see
+ * {@link PerpsInternalTransferType}.
+ *
+ * @experimental This API may change in a breaking way in any release, including patch releases.
+ */
+export enum PerpsKnownInternalTransferType {
+  Transfer = 'transfer',
+  ReferralPayout = 'referral_payout',
+}
+
+/**
+ * An internal-transfer classification. Known values are enumerated in
+ * {@link PerpsKnownInternalTransferType}; newly introduced values flow through
+ * as plain strings so they can be handled before a client release that
+ * enumerates them.
+ *
+ * @experimental This API may change in a breaking way in any release, including patch releases.
+ */
+export type PerpsInternalTransferType =
+  | PerpsKnownInternalTransferType
+  | (string & {});
+
+/**
  * @experimental This API may change in a breaking way in any release, including patch releases.
  */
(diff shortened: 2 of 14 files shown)


osr21:
One recovery detail worth making explicit for callers: the one-attempt/unknown-outcome contract is safe, but after a timeout the action cannot return the new `transferId`, so consumers need a documented correlation strategy before they can decide whether manual intervention is required.

A practical recipe using the surface in this PR would be:

1. Generate a unique caller correlation value (32 hex chars works well) and put it in the signed `label` before calling `transferPerpsCollateral`.
2. Record the local send window, amount, direction/counterparty and label.
3. On an unknown outcome, **do not resubmit**. Poll `listInternalTransfers({ start, end })` over a bounded window and match the full tuple, primarily the unique label.
4. If one match appears, treat its `transferId` as the recovered acknowledgement; if zero/multiple matches remain after the window, surface an explicit uncertain state for operator review.

This matters because amount + counterparty alone are not unique for treasury/rebalancing workloads, while the timestamp-only pagination ceiling noted here makes an unbounded history scan unreliable. If labels are not guaranteed unique server-side, the docs should say that clearly and present the tuple match as reconciliation rather than idempotency. A short example/TSDoc block would make the otherwise good “never retry unknown outcomes” guidance actionable.

naruto11eth (approved):
approved.

osr21 (approved):

PR #332, opened 2026-08-30
feat(data): migrate builder analytics to data api v2

skanda-polymarket:
## Summary
- Migrate builder leaderboard to v2 opaque cursors with normalized builder identities and share volume.
- Migrate builder volume to complete interval buckets with an explicit `bucketLimit`.
- Add v1/v2-backed bindings coverage, a live integration workflow, a runnable example, and a DEV-612 changeset.

## Validation
- `pnpm lint`
- `pnpm typecheck`
- `pnpm build && pnpm test`
- `pnpm exec vitest packages/client/tests/integration/leaderboards.test.ts --project client-integration`
- `pnpm --filter @polymarket/examples analytics:builders`

---

> [!NOTE]
> **Medium Risk**
> Public SDK breaking changes (types, request shapes, pagination semantics) will require downstream updates; scope is read-only analytics, not auth or trading paths.
> 
> **Overview**
> **Breaking** migration of `listBuilderLeaderboard` and `fetchBuilderVolume` from Data API v1 to v2 (`/v2/builders/leaderboard` and `/v2/builders/volume`).
> 
> Builder leaderboard pagination now uses **server opaque cursors** (via `dataPageSchema`) instead of client-side offset encoding; request options use `window` (`LeaderboardWindow`) instead of `timePeriod`, with higher default/max `pageSize` (100 / 1000). Responses are **`BuilderStanding`** rows (`builderCode`, numeric `rank`, share `volume` as string, etc.).
> 
> Builder volume is a time series of **`BuilderVolumePoint`** rows grouped by **`bucketDate`**, with `interval` (`BuilderVolumeInterval`) and **`bucketLimit`** (complete buckets, not row count) replacing `timePeriod`. Bindings add Zod normalization (empty profile image stripped); both calls wrap requests with **rate-limit retry** and use `toDataSearchParams`.
> 
> Trader leaderboard stays on v1. Adds binding unit tests, expanded integration tests (v2 query params and cursor walk), an `analytics:builders` example script, and a minor changeset for `@polymarket/bindings` / `@polymarket/client`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3cb2ca436f8664387d3728c62c50a7bece25990. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ExoMonk:
Reviewed with the live spec and prod probes; the contract side is fully verified: `time_period` vocabulary and wire key, the 1000 limit cap (the server 400s 1001, so the client max mirrors it exactly), `bucketLimit` correctly modeled as buckets rather than rows (cap 90, `all` = yearly), the newest-bucket-first claim (probed: 2026-08-31 then 2026-08-30), and the cursor-pins-window comment matches the boards' cursor-cohort contract. Type conventions check out too: `volume` normalizes to `DecimalString`, `bucketDate` is `IsoCalendarDateString`, integers stay integers.

Two items before merge:

1. **The cursor-walk integration test is deterministically brittle.** It walks two pages of `LeaderboardWindow.Day`, and the day board is empty right after UTC midnight (probed live at 00:20Z: 0 rows, `has_more: false`). Any CI run in the early UTC hours fails on `expectNonEmptyPage`. Suggest walking `LeaderboardWindow.All` for the pagination assertions (always populated) and keeping `Day` only for the query-param assertion.
2. **Rebase needed**: the branch is based on the integration state before #325 merged (DecimalString sweep + enum additions). No file overlap expected, so it should rebase clean, but the gates are worth re-running after.

Gates on the stack ran clean here otherwise: builds, typecheck, lint, bindings tests, and the builder-volume integration test all pass.

ExoMonk:
Re-verified on efa5247/a3cb2ca: the cursor walk now rides the All window with a separate Day param-forwarding test (exactly the fix), and the branch is synced with current integration. Full gates re-run on the stack: builds, typecheck, lint, bindings 131/131, and both integration suites pass live. LGTM from the contract side.

ExoMonk (approved):

packages/bindings/src/data/leaderboard.ts
```diff
@@ -1,33 +1,116 @@
...
-export const BuilderVolumeEntrySchema = z
+export const LeaderboardWindowSchema = z.enum(LeaderboardWindow);
+
+/** Bucket width used by the builder volume time series. */
+export enum BuilderVolumeInterval {
+  Day = 'day',
+  Week = 'week',
+  Month = 'month',
+  /** Groups the series by calendar year. */
+  All = 'all',
```
cesarenaldi, line 29: The `all` interval appears to return calendar-year buckets, so I’m wondering whether `BuilderVolumeInterval.Year = 'all'` would be clearer at the SDK level. I understand that `all` is the current API wire value, which we can preserve in the mapping and potentially align at the API level later. What do you think?

skanda-polymarket: agreed. follow up incoming

skanda-polymarket: Addressed in #340: the SDK now exposes `BuilderVolumeInterval.Year` while preserving the wire value `all`.

PR #333, opened 2026-08-31
feat(data): add resolution lifecycle queries

skanda-polymarket:
## Summary
- add `fetchResolutions` with mutually exclusive question, condition, and event selectors
- normalize lifecycle, oracle, payout, and finality fields, including Gamma-compatible bytes31 condition IDs
- stack this change on #332

## Validation
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`
- `pnpm test:client:integration packages/client/tests/integration/events.test.ts`

---

> [!NOTE]
> **Medium Risk**
> Read-only data API, but resolution rows drive payout/oracle semantics and condition-ID normalization could mis-route lookups if wrong.
> 
> **Overview**
> Adds **`fetchResolutions`** so SDK consumers can load market resolution lifecycle rows from **`GET /v2/resolutions`**, with exactly one of **question**, **condition** (up to 20 IDs), or **event** (up to 20 IDs) as the selector.
> 
> **`@polymarket/bindings`** introduces **`Resolution`** types (status, oracle prices, payouts, reporter/source, finality) and Zod parsing that maps snake_case wire fields to camelCase and **drops unset sentinels** (empty tx/log fields, oracle price **`'69'`**).
> 
> The client validates inputs (deduped ID lists, 32-bit event IDs), **right-pads 31-byte protocol v2 market condition IDs** to 32-byte form for the API, and **rejects 31-byte combo condition IDs**. The call uses **`toDataSearchParams`**, rate-limit retry, and is exposed on **`client.fetchResolutions`** via the data decorator, with resolution enums re-exported from the client.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a13dc951ab5bfd7a7abd356f06317a4668082262. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ExoMonk:
Reviewed against the live spec plus prod probes; this one is in strong shape and two of its choices caught real wire subtleties:

- **The 31-byte condition padding is verified end to end.** v2 structural condition ids genuinely ride the wire as 31 bytes (combo leg payloads serve them), the service 400s them unpadded, and the padded form returns real rows: probed a resolved leg id and got its resolution back through the padding. Good catch, this is exactly the ergonomic bridge consumers need.
- **`MixedDateTimeString` is right and the spec prose is wrong**: the published field doc says condition-keyed rows serve RFC3339, but UMA-sourced condition rows serve epoch-seconds strings (probed `'1764093873'` on a condition lookup). Your type absorbs it; we will fix the spec doc text service-side.
- Absent-vs-null verified on both row shapes (unresolved UMA row and resolved native row): unset fields are absent, never `null`, so `.optional()` is correct for this wire. The `'69'` sentinel strip and empty `transaction_hash`/`log_index` handling match the documented conventions, and the selector union is exactly the agreed exactly-one shape with the 20-id dedupe caps.

One convention item to align before merge: **`proposedPrice`, `reproposedPrice`, and `price` should be `DecimalString`** (via `DecimalStringSchema`) rather than plain `string`. They are prices, which is exactly the category the SDK-surface decimal convention covers (see the #325 thread); the sentinel-strip logic stays as is, only the brand changes. `payouts` staying `number[]` is right (integer micro units).

Gates on the stack ran clean here: builds, typecheck, lint, bindings tests 128/128, and all three resolutions integration tests pass live, including the padding and empty-selector cases. Note the stack needs the #332 rebase onto current integration first.

ExoMonk:
Re-verified on a13dc951a: the three price fields now carry DecimalString, and the boundary hardening is a genuine improvement beyond the ask, so flagging it as verified: the padding now applies only to 31-byte protocol v2 MARKET prefixes and rejects 31-byte combo ids. Cross-checked against the indexer's identity module (module byte 1 = Binary, 2 = Negative-Risk, 3 = Combinatorial): the 0x01/0x02-pad, 0x03-reject split is exactly right, since /v2/resolutions serves market conditions and a padded combo id would silently return an empty answer instead of a typed error. Full gates re-run on the stack: builds, typecheck, lint, bindings 131/131, all resolutions integration tests pass live. LGTM.

ExoMonk (approved):

packages/bindings/src/data/resolutions.ts
```diff
@@ -0,0 +1,167 @@
...
+  /** Second proposed oracle price, omitted until set. */
+  reproposedPrice?: DecimalString;
+  /** Final oracle settlement price, omitted until set. */
+  price?: DecimalString;
+  /** Transaction for the latest lifecycle event, when available. */
+  transactionHash?: string;
+  /** Log index for `transactionHash`, when available. */
+  logIndex?: string;
+  /** Latest lifecycle change, represented as epoch seconds or RFC3339 UTC. */
+  lastUpdatedAt: MixedDateTimeString;
```
cesarenaldi, line 80: `lastUpdatedAt` should expose one SDK datetime representation rather than carrying the wire inconsistency into `MixedDateTimeString`. Please parse `last_update_timestamp` with `EpochLikeToIsoDateTimeStringSchema` and change the public field type to `IsoDateTimeString`. The existing schema handles both epoch-second strings and RFC3339 values, so callers consistently receive:

```ts
// wire
last_update_timestamp: '1785355919'

// SDK
lastUpdatedAt: '2026-07-29T20:11:59.000Z'
```

cesarenaldi (commented):
Timestamp normalization

packages/bindings/src/data/resolutions.ts
```diff
@@ -0,0 +1,167 @@
...
+  /** First proposed oracle price, omitted until set. */
+  proposedPrice?: DecimalString;
+  /** Second proposed oracle price, omitted until set. */
+  reproposedPrice?: DecimalString;
+  /** Final oracle settlement price, omitted until set. */
+  price?: DecimalString;
+  /** Transaction for the latest lifecycle event, when available. */
+  transactionHash?: string;
+  /** Log index for `transactionHash`, when available. */
+  logIndex?: string;
```
cesarenaldi, line 78: Please use the established SDK types for this transaction metadata: expose `transactionHash?: TxHash`, validate non-empty values with `TxHashSchema`, and parse the numeric `log_index` wire string into `logIndex?: number`. Are we sure we need to send `logIndex` as a string? is this an hex quantity?

cesarenaldi (commented):
Transaction metadata normalization

packages/bindings/src/data/resolutions.ts
```diff
@@ -0,0 +1,167 @@
...
+  /** Latest lifecycle change, represented as epoch seconds or RFC3339 UTC. */
+  lastUpdatedAt: MixedDateTimeString;
+  marketType?: ResolutionMarketType;
+  /** Per-outcome payout in micro collateral units per share. */
+  payouts?: number[];
```
cesarenaldi, line 83: Please normalize these micro-USDC values at the binding boundary and expose `payouts` as a two-item `DecimalString` tuple. The backend can continue returning numbers; the SDK should remove the fixed-point scale:

```ts
const PayoutSchema = z
  .number()
  .int()
  .min(0)
  .transform(String)
  .pipe(E6BigIntStringToDecimalStringSchema);

payouts: z.tuple([PayoutSchema, PayoutSchema]).optional();
```

This gives callers `[500_000, 500_000]` as `['0.5', '0.5']`, consistent with SDK monetary normalization, while the tuple also captures the documented `[outcome0, outcome1]` shape.

cesarenaldi (commented):
Payout normalization

packages/bindings/src/data/resolutions.ts
cesarenaldi: Any reason on why the scaled of these are not consistent with the rest of the number values?

PR #336, opened 2026-08-31
fix(client): stop revocation after registry removal

kartojal:
Return from session-key revocation once the key leaves the wallet registry, without waiting for on-chain confirmation.

---

> [!NOTE]
> **Medium Risk**
> Breaking public API and changed completion semantics for session-key revocation; behavior still gates on registry removal so the key should be unusable before return, but callers no longer get on-chain confirmation.
> 
> **Overview**
> **`revokeSessionKey`** no longer blocks on relayer/on-chain confirmation. After the revocation is accepted by the relayer, it **polls active session keys** (reusing relayer poll limits, ~200s) until the address disappears from the registry, then resolves.
> 
> This is a **breaking API change** (minor bump): the method returns **`Promise<void>`**, **`RevokeSessionKeyResult`** is removed, and docs/examples no longer promise a confirmed transaction. Integration tests assert registry state instead of a revocation transaction hash.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3dd560b6cf7dc74f940162502b872ac36bf90ad9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/actions/session-keys.ts
+++ b/packages/client/src/actions/session-keys.ts
@@ -417,6 +417,8 @@ export async function revokeSessionKey(
     kind: 'revocation',
     status: response.status,
   });
+  // This eager return is intentional; the backend continues the remaining
+  // revocation work after the key leaves the active-key registry.
   await waitForRevokedSessionKey(client, parsedRequest.address);
 }
 
@@ -424,6 +426,9 @@ async function waitForRevokedSessionKey(
   client: BaseSecureClient,
   address: EvmAddress,
 ): Promise<void> {
+  // This revocation readiness check deliberately reuses the relayer transaction
+  // polling limits instead of introducing revocation-specific configuration.
+  // The defaults allow about 200 seconds (100 polls every two seconds).
   for (
     let pollCount = 0;
     pollCount < client.environment.relayerMaxPolls;
--- a/packages/client/src/actions/session-keys.ts
+++ b/packages/client/src/actions/session-keys.ts
@@ -35,7 +35,7 @@ import {
 } from '../errors';
 import { parseUserInput } from '../input';
 import { validateWith } from '../response';
-import type { TransactionOutcome } from '../types';
+import type { TransactionHandle, TransactionOutcome } from '../types';
 import { SignerType } from '../wallet';
 import { completeWith } from '../workflow';
 import {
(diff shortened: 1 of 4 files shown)


packages/client/src/actions/session-keys.ts
```diff
@@ -423,14 +417,32 @@ export async function revokeSessionKey(
     kind: 'revocation',
     status: response.status,
   });
-  const transaction = await new GaslessTransactionHandle(client, {
-    transactionHash: null,
-    transactionId: response.transactionId,
-  }).wait();
+  await waitForRevokedSessionKey(client, parsedRequest.address);
```
cesarenaldi, line 422: This looks like a deliberate choice to return eagerly to the caller once the key is no longer listed, rather than wait for the full backend flow. That makes sense because the backend will continue processing the revocation on its own: it waits through the 65-second cache window, cancels the key's orders, and then submits and confirms the on-chain revoke. The trade-off is that the returned promise now means "the key is no longer listed", not "the on-chain revocation is confirmed".

I think we need to align a couple of things:

- Add a short inline comment above this call explaining that the eager return is intentional and that the backend continues the remaining work.
- Add a note on `waitForRevokedSessionKey`, like the one on the authorization helper, explaining that it reuses `relayerMaxPolls` and `relayerPollFrequencyMs` instead of adding revocation-specific settings. With the defaults, the SDK polls for roughly 200 seconds at most: 100 polls at two-second intervals.
- Update the TypeScript SDK tab in the session-key docs to explain the new return point, the backend's 65-second cache window, and the SDK's configurable polling window.

packages/client/src/actions/session-keys.ts
```diff
@@ -365,8 +359,8 @@ export const RevokeSessionKeyError = makeErrorGuard(
 /**
  * Revokes a session key authorized for the Deposit Wallet.
  *
- * Revocation may take several minutes while existing activity is canceled and
- * the on-chain revocation is confirmed.
+ * Returns once the session key is removed from the active-key registry and can
```
cesarenaldi, line 362: Since `revokeSessionKey()` now returns `Promise<void>`, could we update this example to call it without assigning `revocation`? The current example implies that the method still returns a result.

kartojal: ack!

packages/client/src/actions/session-keys.ts
kartojal: gotcha

cesarenaldi (approved):

PR #339, opened 2026-09-01
feat(data): migrate portfolio summaries to v2

skanda-polymarket:
## Summary

- migrate `fetchPortfolioValue` to the V2 object contract and canonical condition filter
- add `fetchUserStats` profile and cumulative PnL models
- normalize fractional money and size fields to `DecimalString`, matching the existing V2 SDK contract
- remove the obsolete `fetchTradedMarketCount` surface while preserving accounting snapshots

Linear: DEV-635

## Validation

- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:types && pnpm test:bindings && pnpm test:client`
- focused live portfolio value and user stats integration tests

---

> [!NOTE]
> **Medium Risk**
> Breaking public SDK surface for portfolio value and traded-market count, with new PnL/stat fields that downstream apps must adopt correctly.
> 
> **Overview**
> **Breaking** portfolio summary APIs move to Data `/v2` with updated shapes and filters.
> 
> `fetchPortfolioValue` now calls `/v2/value` (with rate-limit retry), returns a single `PortfolioValue` (`wallet` + USDC `value` as `DecimalString`) instead of a `Value[]`, and filters via optional `conditionIds` (serialized as `condition`, max 20) rather than legacy `market`. Bindings replace `ValueSchema` with `PortfolioValueSchema` wrapped in `dataEnvelopeSchema`.
> 
> `fetchTradedMarketCount` and `TradedSchema` are removed. **`fetchUserStats`** (`/v2/user-stats`) returns `UserStats | null` with `tradedMarketCount`, profile fields, and optional `allTimePnl` (`UserPnlPoint` with decimal-string money/size fields). Account decorators and error guards are updated accordingly; **`downloadAccountingSnapshot`** is unchanged.
> 
> Unit tests cover schema normalization; integration tests assert v2 query params and response types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3c3f7ff0074a418484ccbb2ba903052fa23ab8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bindings/src/data/portfolio.ts
+++ b/packages/bindings/src/data/portfolio.ts
@@ -242,13 +242,13 @@ export const PositionSchema = z
 export type PortfolioValue = {
   wallet: EvmAddress;
   /** Total marked portfolio value, in USDC, rounded to four decimals. */
-  value: number;
+  value: DecimalString;
 };
 
 export const PortfolioValueSchema = z
   .object({
     proxy_wallet: EvmAddressSchema,
-    value: z.number(),
+    value: DecimalishSchema,
   })
   .transform(({ proxy_wallet, value }) => ({
     wallet: proxy_wallet,
@@ -256,8 +256,8 @@ export const PortfolioValueSchema = z
   })) satisfies z.ZodType<PortfolioValue>;
 
 /**
- * One cumulative wallet PnL observation. Monetary values are exposed as
- * JavaScript numbers with up to six decimal places.
+ * One cumulative wallet PnL observation. Fractional money and size values
+ * are normalized to decimal strings.
  */
 export type UserPnlPoint = {
   /** Observation time as Unix epoch milliseconds. */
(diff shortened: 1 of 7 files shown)


ExoMonk:
Reviewed against the live spec plus prod probes, same drill as #332/#333. The wire-contract side is excellent:

- `UserPnlPoint` models all 27 wire fields with the exact required/nullable split the spec declares; probed a real wallet and the explicit `null`s (`deposits`/`withdrawals`/`cashflow_net`) parse cleanly through the nullish handling.
- `join_date` as epoch-seconds-or-null, `biggest_win` semantics, and the unknown-wallet `data: null` arm all match the live wire.
- `/v2/value`: the `condition` cap (20) mirrors the service, the canonical-query-key test is a nice touch, and the combos-excluded-when-filtered doc matches the value definition (single-market holdings only under a filter; verified live: 671,483 full vs 400,000 filtered).
- The `/traded` retirement is complete: zero dangling references to `Traded`/`fetchTradedMarketCount` anywhere, accounting snapshot carve-out preserved, and the `DefaultAccountWallet` indexed-access cleanup is a conventions win.
- Gates on the branch: builds, typecheck, lint, bindings 132/132.

One convention item to align, and it is the big one: **the money fields should be `DecimalString`, not `number`**. `PortfolioValue.value`, `biggestWin`, and every non-integer monetary field on `UserPnlPoint` (the PnL components, fee family, volume, volumeUsdc) fall under the SDK-surface decimal convention decided on #325 and just applied to #333's price fields: wire numbers normalize through `DecimalishSchema` to `DecimalString` so the whole v2 surface reads evenly. Integers stay integers (`trades`, `views`, `sourceBlock`, `tradeCount`, timestamps). The 'rejects decimal strings' test inverts under the tolerant schema, and the changeset's 'numeric' wording follows the change.

Separately, NOT this PR's doing: the pre-existing `listComboPositions` integration test on the branch fails against the live wire because `realizedPayoutUsdc` expects `Any<Number>` while the field became `DecimalString` in the #325 sweep (branch CI does not run the integration project yet, so it slipped). Since this PR already touches `portfolio.test.ts`, feel free to ride the one-line fix (`realizedPayoutUsdc: expect.any(String)`) here, or leave it and we will patch it on the branch.

packages/client/src/decorators/account.ts
```diff
@@ -149,23 +149,25 @@ export type PublicAccountActions = Prettify<{
    * });
    * ```
    */
-  fetchPortfolioValue(request: FetchPortfolioValueRequest): Promise<Value[]>;
+  fetchPortfolioValue(
+    request: FetchPortfolioValueRequest,
+  ): Promise<PortfolioValue>;
```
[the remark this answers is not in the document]
skanda-polymarket: Addressed in c686bbf: both public and secure fetchPortfolioValue TSDoc now document the 20-distinct-condition cap and that a condition filter excludes portfolio-level combo positions.

skanda-polymarket:
Aligned the second review point with merged #325 in c686bbf: wire numbers or strings normalize through DecimalishSchema, and the SDK returns DecimalString for portfolio value, biggest win, and all fractional money, size, and PnL fields. Counts, blocks, and timestamps remain numbers. Tests and the changeset were updated accordingly.

ExoMonk (approved):

packages/client/src/actions/portfolio.ts
```diff
@@ -374,8 +374,10 @@ export function listComboPositions(
 }
 
 const FetchPortfolioValueRequestSchema = z.object({
-  user: z.string(),
-  market: z.array(z.string()).optional(),
+  user: EvmAddressSchema,
+  conditionId: z
```
cesarenaldi: `conditionId` currently has two shapes at the call site: a single string or an array. `fetchResolutions` already establishes `conditionIds` for batched condition selectors, and I raised the same point in [PR #342](https://github.com/Polymarket/ts-sdk/pull/342#discussion_r3912549170). Please use the same shape here, ideally `conditionIds?: string[]`, so these APIs compose consistently and callers do not write the surprising `conditionId: [a, b]`.

cesarenaldi (commented):
​

packages/bindings/src/data/portfolio.ts
```diff
@@ -239,10 +239,192 @@ export const PositionSchema = z
...
+      cashflowNet: cashflow_net,
+      volumeUsdc: volume_usdc,
+      tradeCount: trade_count,
+    }),
+  ) satisfies z.ZodType<UserPnlPoint>;
+
+export type UserStats = {
+  wallet: EvmAddress;
+  /** Exact number of distinct markets the wallet has traded. */
+  trades: number;
```
cesarenaldi: Am I understanding correctly that `trades` is the number of distinct markets the wallet has traded, rather than the number of trades? If so, this seems misleading at the API level, especially beside `all_time_pnl.trade_count`, which counts fill events. Should the API call this `traded_market_count` instead? At minimum, I think the SDK should expose `tradedMarketCount`.

cesarenaldi (commented):
​

packages/bindings/src/data/portfolio.ts
skanda-polymarket: this was probably inherited from v1's `trades`. I will make the SDK expose `tradedMarketCount`, and I will track the API change separately.

packages/client/src/actions/portfolio.ts
skanda-polymarket: addressed this in https://github.com/Polymarket/ts-sdk/pull/339/changes/085ea41408b7fca99e6713932146cd797f6a8409

packages/bindings/src/data/portfolio.ts
skanda-polymarket: added this in [e3c3f7f](https://github.com/Polymarket/ts-sdk/pull/339/commits/e3c3f7ff0074a418484ccbb2ba903052fa23ab8e), and will track this internally on api

cesarenaldi (approved):

PR #341, opened 2026-09-01
feat(bindings): expose market protocol versions

kartojal:
Expose known market protocol versions while preserving unknown future values as strings.

---

> [!NOTE]
> **Low Risk**
> Additive bindings/schema fields with minor package bumps; stricter parsing only affects responses that include unsupported protocol version strings.
> 
> **Overview**
> Adds a **`ProtocolVersion`** enum (`v1` / `v2`) and optional **`version`** on normalized **markets** and **events**, parsed from Gamma responses and passed through normalization.
> 
> Validation only accepts known enum values or **`null`**; unknown strings (e.g. `v3`) fail schema parsing. **`@polymarket/client`** re-exports **`ProtocolVersion`** for app use.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53a5caa5c700a5cd8c35fb5747405d7055dcd861. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bindings/src/gamma/market.test.ts
+++ b/packages/bindings/src/gamma/market.test.ts
@@ -4,6 +4,7 @@ import {
   ListMarketsKeysetResponseSchema,
   ListMarketsResponseSchema,
   MarketSchema,
+  ProtocolVersion,
 } from './market';
 
 const rawBinaryMarket = {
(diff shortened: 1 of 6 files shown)


packages/bindings/src/gamma/market.ts
```diff
@@ -197,6 +208,8 @@ export type MarketTag = {
 
 export type Market = {
   id: MarketId;
+  /** Protocol version used to trade this market. */
+  version?: MarketProtocolVersion | null;
```
cesarenaldi: Gamma also returns `version` on events. Should this PR add it to `GammaEventSchema` and the normalized `Event` as well?

packages/bindings/src/gamma/market.ts
```diff
@@ -83,6 +83,17 @@ export type ComboStatus = ComboKnownStatus | (string & {});
...
+export enum ProtocolVersion {
+  V1 = 'v1',
+  V2 = 'v2',
+}
+
+export type MarketProtocolVersion = ProtocolVersion | (string & {});
+
+export const ProtocolVersionSchema = z
+  .string()
+  .transform((value): MarketProtocolVersion => value);
```
cesarenaldi: Gamma defines this as a closed `v1 | v2` contract and says unknown values are unsupported. Can we use `z.enum(ProtocolVersion)` here instead of passing arbitrary strings through?

packages/bindings/src/gamma/market.test.ts
```diff
@@ -62,38 +63,49 @@ describe('MarketSchema', () => {
   it.each([
     {
       name: 'legacy market',
```
cesarenaldi: Could we call this a `CTF market` instead of a `legacy market`?

packages/bindings/src/gamma/market.ts
```diff
@@ -83,6 +83,17 @@ export type ComboStatus = ComboKnownStatus | (string & {});
...
 const ComboStatusSchema = z.string().transform((value): ComboStatus => value);
 
+export enum ProtocolVersion {
+  V1 = 'v1',
+  V2 = 'v2',
+}
+
+export type MarketProtocolVersion = ProtocolVersion | (string & {});
+
+export const ProtocolVersionSchema = z
```
cesarenaldi: `ProtocolVersionSchema` is only used to parse a response field, so I don’t think it needs to be exported.

packages/bindings/src/gamma/market.ts
cesarenaldi: I genuinely don't know when Gamma populates it, can an event have a mix and match of CTF and Poly v2 based markets? IDK

kartojal: as today Gamma events cannot mix CTF and PolyV2

packages/bindings/src/gamma/market.ts
kartojal: ok!

packages/bindings/src/gamma/market.test.ts
kartojal: sure!

cesarenaldi (approved):

PR #342, opened 2026-09-01
feat(data): migrate market metrics to v2

skanda-polymarket:
## Summary

- migrate open-interest and live-volume reads to their v2 contracts
- replace `listOpenInterest` with direct `fetchOpenInterest` and validate canonical selectors
- expose taker-volume fields and normalize unidentified condition buckets to `null`

Linear: DEV-636

## Validation

- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:bindings` and `pnpm test:client`
- focused live market and event integration tests

---

> [!NOTE]
> **Medium Risk**
> Public breaking API renames and response shape changes for analytics consumers; behavior is scoped to data reads with stricter input validation rather than auth or funds flow.
> 
> **Overview**
> **Breaking SDK change** for market analytics: open interest and event live volume now call the data service **v2** routes and use updated request/response shapes.
> 
> `listOpenInterest` is removed in favor of **`fetchOpenInterest`**, which hits `/v2/oi` with optional **`conditionIds`** (up to 20 canonical market IDs) and returns **USDC gross OI** rows keyed by `conditionId` (`null` for the global aggregate). Live volume moves to **`/v2/live-volume`** with **`eventIds`** instead of a single `id`; the client now returns one **`LiveVolume`** object with **`takerVolumeTotal`** and per-market **`takerVolume`** (shares), not an array with `total`.
> 
> Bindings parse the v2 wire format (`condition_id`, `GLOBAL`, `dataEnvelopeSchema`) and drop deprecated `market`/`value` live-volume fields. Shared **`CanonicalMarketConditionIdSchema`** and **`PositiveInt32EventIdSchema`** centralize selector validation (also used for resolutions). Both actions use **`toDataSearchParams`**, **`withRateLimitRetry`**, and updated integration tests assert the v2 paths and query params.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0612a64c5f29257bdae8d8afc3b1b50791ede01f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bindings/src/data/analytics.ts
+++ b/packages/bindings/src/data/analytics.ts
@@ -69,26 +69,20 @@ export const MetaHolderSchema = z
   })) satisfies z.ZodType<MetaHolder>;
 
 export type OpenInterest = {
-  /** Condition ID of the market, or `GLOBAL` for the global aggregate. */
-  conditionId: ConditionId | 'GLOBAL';
+  /** Condition ID of the market, or `null` for the global aggregate. */
+  conditionId: ConditionId | null;
   /** Priced gross open interest in USDC. */
   value: DecimalString;
 };
 
 export const OpenInterestSchema = z
   .object({
-    condition_id: z.union([
-      ConditionIdSchema.refine(
-        (conditionId) => conditionId.length === 66,
-        'Expected a 32-byte condition ID',
-      ),
-      z.literal('GLOBAL'),
-    ]),
+    condition_id: z.union([ConditionIdSchema, z.literal('GLOBAL')]),
     value: DecimalishSchema,
   })
-  .transform(({ condition_id, ...rest }) => ({
-    ...rest,
-    conditionId: condition_id,
+  .transform(({ condition_id, value }) => ({
+    conditionId: condition_id === 'GLOBAL' ? null : condition_id,
+    value,
   })) satisfies z.ZodType<OpenInterest>;
 
 export type MarketLiveVolume = {
@@ -100,13 +94,7 @@ export type MarketLiveVolume = {
 
 export const MarketLiveVolumeSchema = z
   .object({
-    condition_id: z.preprocess(
-      emptyStringToNull,
-      ConditionIdSchema.refine(
-        (conditionId) => conditionId.length === 66,
-        'Expected a 32-byte condition ID',
-      ).nullable(),
-    ),
+    condition_id: z.preprocess(emptyStringToNull, ConditionIdSchema.nullable()),
     taker_volume: DecimalishSchema,
   })
   .transform(({ condition_id, taker_volume }) => ({
--- a/packages/bindings/src/data/analytics.ts
+++ b/packages/bindings/src/data/analytics.ts
@@ -11,6 +11,7 @@ import {
   type PositionId,
   type TokenId,
 } from '../shared';
+import { dataEnvelopeSchema } from './envelope';
 
 export type Holder = {
   wallet: EvmAddress | null | undefined;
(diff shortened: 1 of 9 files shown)


ExoMonk:
Reviewed with the usual discipline: full diff read against the live OpenAPI (`/v2/openapi.json`), prod probes on both endpoints, and the full gate chain on the PR head (`d64f383`) in a fresh worktree. **LGTM, no findings.** Everything checked out, including the sharp edges:

**Verified against prod:**

1. **Canonical id handling is exactly right.** The service 400s raw 62-hex structural ids (`{"error":"invalid condition id: 0x01304436b57e87850ca3d1ab20706670060000000000000000000000000000"}`) and accepts the padded 64-hex form, so the pad-to-canonical transform is required, not cosmetic. A padded combo id (`0x030b98f62eadaee25b8c04abe48fb41d48000000000000000000000000000000`) returns `200 {"data":[]}` (absent = not servable), so the client-side `0x03` rejection with a typed message is a strictly friendlier version of the same outcome, consistent with the #333 convention.
2. **The `length === 66` response refines are safe.** `/v2/oi` echoes the canonicalized requested id (probed active market `0x5e464d85eb49f22d876f3ed6168a7db5e2288e9ae1eb91effd2758e994676f86`, echo intact, both `condition` and the `conditionId` alias). `/v2/live-volume` serves the market's real 32-byte on-chain condition id even for protocol-v2 markets (probed event 738032: `0x435620fa180dbedb59d34f164d82a447304436b57e87850ca3d1ab2070667006`, 66 chars).
3. **The `""` to `null` normalization is load-bearing, not defensive fiction.** Prod really serves it: `GET /v2/live-volume?event_id=603507` returns `{"data":{"taker_volume_total":961701.098775,"conditions":[{"condition_id":"","taker_volume":961701.098775}]}}`. Without the `emptyStringToNull` preprocess the whole response would fail validation for these events. Good call.
4. **Params match the wire contract**: `condition` CSV capped at 20 distinct (dedupe runs on the canonical lowercased form, so mixed 62-hex/padded spellings of one market dedupe to one entry), `event_id` CSV with no cap (spec documents none; unparseable ids are only a service 400 when all are bad, and the stricter client-side reject is the house pattern). `toDataSearchParams` comma-joins arrays as the spec wants.
5. **Cesare's convention holds**: `value` and `taker_volume`/`takerVolumeTotal` land as `DecimalString` on the surface (wire is JSON number, `DecimalishSchema` normalizes); ids stay strings; doc comments state units (USDC / shares) and the zero-vs-absent semantics verbatim from the spec.

**Gates** (worktree at `d64f383`, 0 behind `integration/data-api-v2`): build, typecheck, lint clean; bindings 131/131; client 404/404.

**Service-side context, not on this PR**: that `""` bucket is a real data bug on our side. Every event owning an empty-key `core.dim_markets` row serves the IDENTICAL global unattributed aggregate (961701.098775 shares) as its entire live volume; probed events 603507, 603541, 604652, 605269 all return the same number, and `/v1/live-volume?id=603507` leaks identically (same view). Filing a DENG ticket to filter `condition_id != ''` in `vw_live_volume` and chase the dim_markets hygiene; the SDK's null handling is the right client posture either way.

ExoMonk (approved):

packages/bindings/src/data/analytics.ts
```diff
@@ -68,58 +69,80 @@ export const MetaHolderSchema = z
...
+  markets: MarketLiveVolume[];
+};
 
-export const LiveVolumeSchema = z.object({
-  total: DecimalishSchema.nullish(),
-  markets: z.array(MarketVolumeSchema).nullish(),
-});
+export const LiveVolumeSchema = z
+  .object({
+    taker_volume_total: DecimalishSchema,
```
cesarenaldi, line 114: My agent spotted that `taker_volume_total` appears to include floating-point arithmetic noise. The individual condition volumes have six-decimal precision, but the aggregated value currently contains additional digits:

```bash
curl -sS 'https://data-api.polymarket.com/v2/live-volume?event_id=106884' \
  | jq '.data.taker_volume_total'
```

```text
4327731.909317001
```

The exact sum at the documented six-decimal precision is `4327731.909317`. Could the API round the aggregate to the same precision before serialization? I do not think the SDK should have to infer and correct the intended precision.

cesarenaldi (commented):
Upstream total-volume precision

packages/bindings/src/data/analytics.ts
```diff
@@ -68,58 +69,80 @@ export const MetaHolderSchema = z
...
 
 export const OpenInterestSchema = z
   .object({
-    market: ConditionIdSchema.nullish(),
-    value: DecimalishSchema.nullish(),
+    condition_id: z.union([
+      ConditionIdSchema.refine(
+        (conditionId) => conditionId.length === 66,
+        'Expected a 32-byte condition ID',
+      ),
```
cesarenaldi: This refinement only restates the API guarantee that response `condition_id` values are 32 bytes. When the API honors its contract it never changes the result, and a non-32-byte response would be an API defect to fix at the source rather than a separate SDK representation. Please use `ConditionIdSchema` directly and remove the same redundant refinement from [`MarketLiveVolumeSchema`](https://github.com/Polymarket/ts-sdk/pull/342/files#diff-c707abfffc3d33df48027bbf3aa83a5274c24fd9bf64d12b5056216b9c98ee45R105) and [`ResolutionSchema`](https://github.com/Polymarket/ts-sdk/pull/333/files#diff-9d2acc5b32615e94eec3ab7e782ff4ef45824e86f21d2de978835059bf8b2394R94).

cesarenaldi (commented):
Deduplicate 32-byte condition response validation

packages/bindings/src/data/analytics.ts
```diff
@@ -68,58 +69,80 @@ export const MetaHolderSchema = z
   })) satisfies z.ZodType<MetaHolder>;
 
 export type OpenInterest = {
-  /** Condition ID of the market whose open interest is reported. */
-  conditionId: ConditionId | null | undefined;
-  /** @deprecated Use `conditionId`. */
-  market: ConditionId | null | undefined;
-  value?: DecimalString | null;
+  /** Condition ID of the market, or `GLOBAL` for the global aggregate. */
+  conditionId: ConditionId | 'GLOBAL';
```
cesarenaldi: I understand that `GLOBAL` is a string in the service’s wire format, but type safety should take precedence at the SDK boundary. `ConditionId | 'GLOBAL'` remains assignable to `string`, so a caller could forward it into another request expecting a condition ID without any compiler error:

```ts
const [{ conditionId }] = await client.fetchOpenInterest();
await client.fetchOpenInterest({ conditionId });
```

If the value is `GLOBAL`, this only fails at runtime. Please normalize `GLOBAL` to `null` and expose:

```ts
conditionId: ConditionId | null;
```

When open interest is not scoped to a specific condition, there is no condition ID. `null` represents that directly and forces callers to narrow the value before reusing it.

cesarenaldi (commented):
Normalize global open interest identity

packages/client/src/actions/markets.ts
```diff
@@ -447,13 +448,26 @@ export async function listMarketHolders(
   );
 }
 
-export type ListOpenInterestError =
+const FetchOpenInterestRequestSchema = z.object({
+  conditionId: z
```
cesarenaldi: `conditionId` currently has two shapes at the call site: a single string or an array. `fetchResolutions` already establishes `conditionIds` for batched condition selectors. Please use the same shape here, ideally `conditionIds?: string[]`, so these APIs compose consistently and callers do not write the surprising `conditionId: [a, b]`.

cesarenaldi (commented):
Align batched condition selector naming

packages/bindings/src/data/analytics.ts
skanda-polymarket: will verify, and standardize this on the api.

packages/bindings/src/data/analytics.ts
skanda-polymarket: agreed. fixing

skanda-polymarket:
[0612a64](https://github.com/Polymarket/ts-sdk/pull/342/commits/0612a64c5f29257bdae8d8afc3b1b50791ede01f) has the following:

- batched `conditionIds` selectors
- `conditionId: ConditionId | null;`

cesarenaldi (approved):

PR #343, opened 2026-09-01
feat(data): add user pnl and volume

skanda-polymarket:
## Summary

- Add `fetchUserPnl` and `fetchUserVolume` with v2 binding schemas, public/secure account methods, and decimal-string normalization.
- Reuse shared rate-limit/window handling and fix the inherited combo payout integration assertion. Stacked on #339.

## Validation

- `pnpm lint` and `pnpm typecheck`
- `pnpm build` and `pnpm test`
- Live portfolio integration: 12 passed, 6 credential-gated skipped

---

> [!NOTE]
> **Low Risk**
> Additive SDK and schema surface with tests; no auth or payment logic changes beyond new read-only data endpoints.
> 
> **Overview**
> Adds **`fetchUserPnl`** and **`fetchUserVolume`** to the data client, wired through v2 **`/v2/user-pnl`** and **`/v2/user-volume`** with the same rate-limit/retry and response validation patterns as other portfolio actions.
> 
> **Bindings** gain `UserPnlInterval` / `UserPnlFidelity`, full **`UserPnlSeries`** and **`UserVolume`** types (snake_case → camelCase, money as decimal strings, nullable PnL fields left null), plus **`FetchUserPnlResponseSchema`** and **`FetchUserVolumeResponseSchema`** with unit tests.
> 
> **Client surface**: low-level actions and **public** / **secure** account methods (secure paths default `user` to the authenticated wallet). PnL accepts optional interval/fidelity; volume accepts the shared optional **`window`** (merged into query params as `start`/`end`). **`UserPnlInterval`** and **`UserPnlFidelity`** are re-exported from the account decorator.
> 
> Integration coverage exercises live responses, canonical volume query keys, and secure wallet defaults. Combo position integration now expects **`realizedPayoutUsdc`** as a string, matching decimal-string normalization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0586ebdafdcf98bfff5fd0940ff9fde2a6b2ecc4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bindings/src/data/portfolio.ts
+++ b/packages/bindings/src/data/portfolio.ts
@@ -255,6 +255,30 @@ export const PortfolioValueSchema = z
     value,
   })) satisfies z.ZodType<PortfolioValue>;
 
+/** Window served by a user PnL series. */
+export enum UserPnlInterval {
+  Max = 'max',
+  All = 'all',
+  OneMonth = '1m',
+  OneWeek = '1w',
+  OneDay = '1d',
+  TwelveHours = '12h',
+  SixHours = '6h',
+}
+
+export const UserPnlIntervalSchema = z.enum(UserPnlInterval);
+
+/** Time step between points in a user PnL series. */
+export enum UserPnlFidelity {
+  OneDay = '1d',
+  EighteenHours = '18h',
+  TwelveHours = '12h',
+  ThreeHours = '3h',
+  OneHour = '1h',
+}
+
+export const UserPnlFidelitySchema = z.enum(UserPnlFidelity);
+
 /**
  * One cumulative wallet PnL observation. Fractional money and size values
  * are normalized to decimal strings.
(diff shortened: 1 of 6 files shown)


ExoMonk:
Same discipline as the last rounds: full delta read against the live OpenAPI, prod probes on both endpoints, service source cross-check, full gate chain on the PR head (`d320dfa`) in a fresh worktree. One minor forward-compat finding below; everything else verified clean.

**Verified against the live spec + prod:**

1. **Enums are exact**: `interval` 7/7 (`max`, `all`, `1m`, `1w`, `1d`, `12h`, `6h`) and `fidelity` 5/5 (`1d`, `18h`, `12h`, `3h`, `1h`) match the published param domains verbatim, and the documented defaults (`1d` / `1h`) are what prod echoes on a bare call (probed `user=0x7c3db723f1d4d8cb9c550095203b686cb11e5c6b`: `interval: "1d", fidelity: "1h", source_fidelity: "1d"`, 24 points).
2. **Wire field names exact**: `proxy_wallet` / `interval` / `fidelity` / `source_fidelity` / `points`, all 27 UserPnlPoint keys present on prod rows; series probed at `1w/3h` (56 points) and `max/1d` (935 points).
3. **user-volume contract matches**: params `user`/`start`/`end` (test asserts the canonical keys on the wire), epoch-second bounds floored to whole UTC days per spec, `{volume, volume_usdc, trade_count}` with shares/USD units documented exactly as the spec pairs them; probed a real window (`{"volume":265162.02507,"volume_usdc":87272.445213,"trade_count":413}`) and an inverted window (zeros, not an error, as the spec promises).
4. **Cesare's convention holds**: all money/size surface fields are DecimalString over JSON-number wire; `tradeCount` stays int. Also confirmed the stacked base picked up `c686bbf` (the #339 decimal alignment) and the inherited `realizedPayoutUsdc` integration assertion fix is in this diff, so the stack lands convention-complete.
5. **Gates** (worktree at `d320dfa`): build, typecheck, lint clean; bindings 134/134; client suite green.

**One minor finding (non-blocking): `sourceFidelity` should not be the closed enum.**

The published contract types `source_fidelity` as a plain string, with prose that exists precisely to signal future grain changes ("Historical MVP observations are daily even when carried onto a finer grid"). The service currently hardcodes it (`crates/service/src/user_pnl.rs`: `const SOURCE_FIDELITY: &str = "1d"`), so the 5-value `UserPnlFidelitySchema` covers today's domain. But unlike `interval` and `fidelity`, which are echoes of closed request-param domains, `source_fidelity` is not documented as closed: the day the substrate gains finer native observations and the service emits a new token (a spec-compatible change on its side, and note a minute-grain token could not even reuse `1m`, which already means one month in the interval domain), every `fetchUserPnl` call would start failing validation with `UnexpectedResponseError`.

Suggestion: parse `source_fidelity` with `z.string()` and surface `sourceFidelity: string` (doc comment can note it is currently always `1d`), keeping the enum only for the two request-echo fields.

packages/bindings/src/data/portfolio.ts
```diff
@@ -255,6 +255,30 @@ export const PortfolioValueSchema = z
     value,
   })) satisfies z.ZodType<PortfolioValue>;
 
+/** Window served by a user PnL series. */
+export enum UserPnlInterval {
+  Max = 'max',
+  All = 'all',
```
cesarenaldi: `Max` and `All` produce the same full-history result, so exposing both gives SDK consumers two choices without a behavioral distinction. Could we keep `Max` as the single canonical option and remove `All` from the public SDK enum?

cesarenaldi (commented):
​

packages/bindings/src/data/portfolio.ts
skanda-polymarket: Done — removed `All`; `Max` is now the only public SDK option for the full-history window.

cesarenaldi (approved):

PR #347, opened 2026-09-02
feat(data): migrate trader boards to v2

skanda-polymarket:
## Summary
- migrate trader leaderboard reads to cursor-paginated v2 and split by-wallet standing into `fetchTraderLeaderboardStanding`
- add `listBiggestWinners` with explicit market and Combo result variants
- normalize identifiers, amounts, timestamps, profile sentinels, and rank absence at the bindings boundary

## Validation
- `pnpm -r --sort --if-present run build`
- `pnpm test:types`, `pnpm test:bindings`, and `pnpm test:client`
- live `client-integration` leaderboard suite (6 tests)
- `pnpm lint`
- `pnpm -r --sort --if-present run typecheck`

---

> [!NOTE]
> **Medium Risk**
> Breaking public SDK types and request parameters for `listTraderLeaderboard` will break downstream callers until they adopt v2 fields and pagination; new data normalization (ranks, amounts, standing nulls) can surface subtle UI bugs if not handled.
> 
> **Overview**
> **Breaking** migration of trader leaderboard reads from legacy `/v1/leaderboard` (offset pagination, `orderBy` / `timePeriod` / fixed category enums) to **`/v2/leaderboard`** with server cursors, `sortBy` (`TraderLeaderboardSort`), `window` (`LeaderboardWindow`), and free-form `category` strings. Response shapes are normalized in bindings: numeric ranks, `wallet` + decimal-string `pnl`/`volume`, nullable profile fields, and paginated `{ items, hasMore, nextCursor }` instead of a bare array.
> 
> Adds **`fetchTraderLeaderboardStanding`** (same v2 route with `user`) for per-wallet PnL/volume and separate `pnlRank` / `volumeRank` (nullable when unranked), exposed on both analytics-style listing and **account** decorators (secure clients default `user` to the authenticated wallet).
> 
> Adds **`listBiggestWinners`** on `/v2/biggest-winners` with cursor pagination and a discriminated **`BiggestWinner`** model (`market` vs `combo`: different IDs, combo rows use `null` event metadata). Removes the old `TimePeriod` / `LeaderboardCategory` / `LeaderboardOrderBy` exports from data common; page defaults rise to 100 (max 1000). Unit and integration tests cover parsing, cursor walks, and combo winner shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d65e0e23c0b487aadd72d1a5c94b2b02e0035864. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/decorators/analytics.ts
+++ b/packages/client/src/decorators/analytics.ts
@@ -6,6 +6,7 @@ import type {
   TraderLeaderboardEntry,
   TraderLeaderboardStanding,
 } from '@polymarket/bindings/data';
+import type { Prettify } from '@polymarket/types';
 import {
   type FetchBuilderVolumeRequest,
   type FetchTraderLeaderboardStandingRequest,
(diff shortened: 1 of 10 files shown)


packages/client/src/decorators/analytics.ts
```diff
@@ -177,17 +238,26 @@ export function analyticsActions(client: BaseClient): AnalyticsActions {
     listBuilderLeaderboard: listBuilderLeaderboard.bind(null, client),
     fetchBuilderVolume: fetchBuilderVolume.bind(null, client),
     listTraderLeaderboard: listTraderLeaderboard.bind(null, client),
+    fetchTraderLeaderboardStanding: fetchTraderLeaderboardStanding.bind(
```
cesarenaldi: `fetchTraderLeaderboardStanding` is wallet-scoped, but the `SecureClient` binding still requires `user`. Please give it the same account binding as `fetchUserStats`:

```ts
await secureClient.fetchTraderLeaderboardStanding({
  window: LeaderboardWindow.All,
});
```

This should default to the authenticated wallet, while the `PublicClient` method continues to require `user`.

cesarenaldi (commented):
​

ExoMonk:
**DEV-637.** Reviewed with the usual discipline: full diff against the live OpenAPI, prod probes, service-source cross-checks, and the full gate chain on the PR head (`7992375e30af65bd70cdbf0fbe48234272b9150d`). **LGTM, no findings.**

**Verified:**

1. **The unranked sentinel is handled exactly right, and I chased it to the source.** The route prose says "a rank of `0` means unranked", but the component schema types `rank_pnl`/`rank_volume` as `integer | null`, and the service maps the substrate's `0` sentinel to `null` at the scan layer (`crates/db/src/clickhouse.rs`: `rank_pnl: (rank_pnl > 0).then_some(rank_pnl)`, with a dedicated test). Probed live: the top combos wallet serves `rank_pnl: 1, rank_volume: null` (the volume board has no combos rows), and the same wallet on overall serves both ranks as integers. So `positive().nullable()` is exactly the wire domain; `0` can never ride it. The stale "rank of 0" prose is a service-side doc bug on our plate, not yours.
2. **Unknown wallet** serves `{"data": null}` (probed) and the nullable envelope returns `null`.
3. **Biggest-winners wire matches the fixtures byte-for-byte** (probed both arms): combo rows carry a 62-hex `0x03` condition id, `event_id: 0`, empty `event_slug`, and the ` / `-joined leg title; market rows carry a numeric `event_id` and a 66-char condition id. The discriminated union with `z.literal(0)` / `z.literal('')` on the combo arm plus the `eventId: null` normalization is the right shape, and forcing the `kind` branch before event metadata is exactly what the spec asks consumers to do.
4. **Cursor pinning**: resending the same `category`/`sort_by`/`time_period` alongside the cursor matches the builders-board pattern and can never trip the 400 (only contradictory restatements do). Test asserts the resend on page 2.
5. **Free-string `category`** replacing the old closed enum is correct: the domain is Gamma categories plus the synthetic `combos`/`esports`, which a closed enum would break on every new category.
6. **Convention**: `pnl`/`volume`/`initialValue`/`finalValue` land as DecimalString over JSON-number wire; ranks stay ints; `resolvedAt` normalizes to epoch millis; `VOLUME` (not v1's `VOL`).

**Gates** at `7992375e30af65bd70cdbf0fbe48234272b9150d`: build, typecheck, lint clean; bindings + client suites green. Cursor Bugbot: 0 findings.

cesarenaldi (approved):

PR #348, opened 2026-09-02
feat(data): migrate price history to v2

skanda-polymarket:
## Summary
- replace legacy fetchPriceHistory with cursor-paginated listPriceHistory on the v2 data route
- expose strict interval, range, and as-of requests with normalized decimal prices and millisecond timestamps

## Validation
- pnpm lint
- pnpm typecheck
- pnpm build
- pnpm test
- targeted live markets integration suite (14 tests)

---

> [!NOTE]
> **Medium Risk**
> Breaking public API and response shape will break integrators on the old CLOB price-history path; validation and pagination behavior are new contract surface area.
> 
> **Overview**
> **Breaking:** removes CLOB `fetchPriceHistory` and adds cursor-paginated **`listPriceHistory`** on the data **`/v2/prices-history`** route. Callers use `assetId` (sent as `token_id`), async pagination, and a stricter request shape.
> 
> Time selection is **exactly one** of: relative `interval` (including new `1m`), explicit `start`/`end` (≤15 days), or snapshot `asOf`. Optional **`bucketSeconds`** replaces `fidelity`, with per-interval minimums enforced client-side before the network call.
> 
> **`PriceHistoryPoint`** is normalized: millisecond `timestamp`, decimal-string `price`, and `resolutionSeconds`. Types and **`PriceHistoryInterval`** move from `@polymarket/bindings/clob` to **`@polymarket/bindings/data`**; the client exposes `listPriceHistory` on data actions and re-exports `PriceHistoryInterval` from markets actions. SDK direction doc now standardizes **`assetId`** for exchange assets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 440544bb1f205b270e56b00ca72b90d23e1304aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/actions/markets.ts
+++ b/packages/client/src/actions/markets.ts
@@ -1,8 +1,10 @@
 import {
   ConditionIdSchema,
   IsoDateTimeStringSchema,
+  type PaginationCursor,
   PaginationCursorSchema,
   PositionIdSchema,
+  TokenIdSchema,
 } from '@polymarket/bindings';
 import {
   type ComboMarket,
@@ -11,8 +13,12 @@ import {
 import {
   FetchOpenInterestResponseSchema,
   ListMarketHoldersResponseSchema,
+  ListPriceHistoryResponseSchema,
   type MetaHolder,
   type OpenInterest,
+  PriceHistoryInterval,
+  PriceHistoryIntervalSchema,
+  type PriceHistoryPoint,
 } from '@polymarket/bindings/data';
 import {
   FetchMarketTagsResponseSchema,
@@ -40,11 +46,14 @@ import { withRateLimitRetry } from '../retry';
 import {
   CanonicalMarketConditionIdSchema,
   distinctIdList,
+  EpochSecondsLikeSchema,
   snakeCase,
   toDataSearchParams,
   toLegacyDataSearchParams,
   toSearchParams,
 } from './params';
+
+export { PriceHistoryInterval };
 
 // The public markets endpoint forces active=true and archived=false server-side.
 const ListMarketsRequestSchema = z.object({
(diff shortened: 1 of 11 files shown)


ExoMonk:
**DEV-636 (prices-history).** Reviewed with the usual discipline: full diff against the live OpenAPI, prod probes on every client-side mirror, gates on the stack head. **LGTM, no findings.**

**Verified against prod** (all probed live on an active token):

1. **The per-interval `bucketSeconds` floors mirror the service exactly**: `interval=1w&bucket_seconds=60` is a service `400 {"error":"bucket_seconds must be at least 300 for this interval"}` and `interval=1m&bucket_seconds=300` is `400 "... at least 600 ..."`, matching the superRefine's 600/300/60 ladder (and the spec's "600 for max/all/1m, 300 for 1w"). Your client-side messages even match the service wording.
2. **The 15-day window mirror is real**: a 16-day explicit window AND a bare `start` 20 days back are both service 400s ("'start' to 'end' must span at most 15 days"), so the client check with `now` as the open-ended upper bound reproduces the service's own rule, not an invented one.
3. **Zero bounds** are service 400s ("must be a positive epoch second"), mirrored by the positive() pipe.
4. **`as_of` serves a page shape**: probed `200` with one point, `resolution_seconds: 0` (exact tick), and `pagination` present, so validating all three arms with the page schema is correct, and "at most one item" in the TSDoc is accurate.
5. **The three-arm union** makes invalid combinations unrepresentable (type tests cover the conflicts), matching the "exactly one window form" contract; `token_id`/`interval`/`bucket_seconds`/`cursor` land snake_case on the wire (test-asserted), and the cursor walk resends the selection.
6. **Convention**: prices land as DecimalString, timestamps as epoch millis, `resolutionSeconds` stays int; pageSize default = cap (10,000) per the spec.

One observation, no action needed: the enum deliberately drops the `all` alias (server treats it as `max`), which fits the one-name-per-behavior bar; anyone porting v1 code will find `Max` by type search.

**Gates** (via the stacked #349 head, which contains this branch, plus this head builds standalone in CI): build, typecheck, lint clean; bindings + client suites green. Cursor Bugbot: 0 findings.

packages/client/src/actions/markets.ts
```diff
@@ -448,6 +457,192 @@ export async function listMarketHolders(
   );
 }
 
+export type ListPriceHistoryRequest =
+  | {
+      tokenId: string;
```
cesarenaldi: Please use `assetId` instead of `tokenId` across all three `ListPriceHistoryRequest` variants. `AssetId` is the SDK’s ledger-agnostic term for either a CTF token ID or a Poly v2 PositionManager position ID. The Data API’s `token_id` name should remain a wire-level detail:

```ts
client.listPriceHistory({
  assetId,
  interval: PriceHistoryInterval.OneDay,
});
```

cesarenaldi (changes requested):

cesarenaldi (approved):

PR #351, opened 2026-09-08
feat(client): add PolyBolt price streams

kartojal:
Add authenticated PolyBolt prices through `client.subscribe()` with production defaults, canonical USD symbols, and fixed 60-second TWAP. Keep RTDS price topics deprecated, with removal planned one month after 0.11.0. Use focused frame tests and the shared integration fixture.

Validation: build, lint, typecheck, and 588 unit/type tests pass. Production integration is blocked during setup because the local Deposit Wallet configuration does not match a deployed production wallet.

---

> [!NOTE]
> **Medium Risk**
> Large new authenticated streaming path with reconnection, multi-socket merging, and a breaking migration story from RTDS feeds; legacy topics remain for now but consumers must adopt new symbols and secure clients.
> 
> **Overview**
> Adds **authenticated PolyBolt live prices** on secure clients through `client.subscribe()` for `prices.crypto`, `prices.crypto.twap` (fixed **60s** TWAP), and `prices.equity`, wired to production `wss://ws-live-v2.polymarket.com/ws`. Subscriptions deliver **history snapshots** plus live updates with **exact decimal strings**, validate **canonical lowercase USD pairs** (e.g. `btcusd`), and surface server rejections as **`RequestRejectedError`** with machine-readable codes.
> 
> Implements a new **realtime WebSocket stack** (shared connection pool, auth, batching, reconnect/backoff, >64-filter fan-out) and **bindings** frame normalization. **`SubscribeError`** now includes connection loss and request rejection; mixed batches **roll back** partial opens on failure. **Legacy RTDS** price topics stay available but are **deprecated** (removal ~1 month after 0.11.0), with docs/changeset calling out USD vs USDT and no 30s TWAP replacement.
> 
> Also adds a **realtime prices example**, integration workflow **`environment_config`**, and metered integration tests for the protocol and transport.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 317b951db3c01571a11aa206bdfe831def48c69d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/client/src/websockets/realtime/socket.ts
```diff
@@ -0,0 +1,457 @@
...
+          key.resolve();
+        }
+      }
+    } catch (error) {
+      if (generation !== this.#generation || this.#closed) return;
+      if (error instanceof SubscriptionRejectedError) this.#fail(error);
+      else if (error instanceof Error) {
+        this.#fail(error);
+      }
+    }
```
samsondav, line 323: Both catch arms call `#fail`, which rejects every key on this socket, ends every listener, and closes the connection. A `bad_filter` or a 10-second ack timeout then kills sibling streams that the server kept. Scope `SubscriptionRejectedError` to the batch keys. Treat ack timeout and send failure as a restart, not a permanent fail-all.

packages/client/src/websockets/realtime/socket.ts
```diff
@@ -0,0 +1,457 @@
...
+        lastResync: Number.NEGATIVE_INFINITY,
+      };
+      this.#keys.set(subscription.key, state);
+      if (this.#authenticated) this.#enqueue('subscribe', [state]);
+    }
+    state.listeners.add(listener);
+    if (state.snapshot !== undefined) listener.event(state.snapshot);
+    if (!this.#started) {
+      this.#started = true;
+      void this.#connect().catch(() => this.#schedule(1006));
```
samsondav: A failed first `#connect()` only schedules reconnect. Nothing rejects `state.ready`. `subscribe()` therefore hangs on DNS failure, a bad URL, or an unreachable host, and the caller never gets a handle to close. Reject never-subscribed keys after the first failed connect (or a deadline), and keep reconnect for keys the server already accepted.

packages/client/src/websockets/realtime/socket.ts
```diff
@@ -0,0 +1,457 @@
...
+    }
```
samsondav, line 130: `add()` reconnects only when `#started` is false. After the last `remove`, a drop in the 1-second idle window skips `#schedule` (`#keys.size === 0`) and leaves `#started === true`. The next `add()` then returns `state.ready` without calling `#connect()`. Connect when the socket is not authenticated and not already connecting, rather than only on the first `add()`.

samsondav (commented):
Upfront: my agent ran this review; flag anything that reads off and I'll take a personal look.

This pull request ships the customer subscribe path for authenticated price streams (`prices.crypto`, `prices.crypto.twap`, `prices.equity`, `prices.polymarket`). The production host flip is still deferred. I request changes for two hangs on the default price socket.

## Requested changes

1. One failed control frame ends every listener on the shared socket.
   `#flush` calls `#fail` on any `SubscriptionRejectedError` or `TransportError`, including the 10-second ack timeout. `#fail` rejects every key, ends every listener, then closes the connection. A `bad_filter` or a single timeout then kills sibling streams that the server kept. Scope the reject to the batch keys. Treat ack timeout and send failure as a restart.

2. `subscribe()` never settles when the first connect fails.
   `add()` does `void this.#connect().catch(() => this.#schedule(1006))`. Retries continue while keys remain. `state.ready` is never rejected. Callers that await `subscribe()` hang and cannot close a handle they never received. Reject never-subscribed keys after the first failed connect, or after a deadline.

A related hang: after the last `remove`, a drop in the 1-second idle window leaves `#started === true` and `#authenticated === false`. The next `add()` does not call `#connect()`.

## Scope

The changeset documents that `wss://ws-live-v2.polymarket.com/ws` is not deployed yet. A merge to `main` publishes through Release and canary. Hold the production default, or hold the merge, until that host is live. This is sequencing, not a logic defect in the client.

`keyTarget` only halves on `dropped` and never recovers. That is a capacity ratchet, not a wrong price.

## Notes

- The branch added `legacy.ts` and then deleted it. That matches the RTDS deletion.
- No prior review threads.
- The changeset tells consumers to remove `protocol`, `rtdsLegacy`, and `includeSnapshot`. Those names were not on the prior public surface.
- The example script requires `POLYMARKET_INTEGRATION_ENVIRONMENT_CONFIG`, so it does not run for an external user on the default production object.

kartojal:
> Upfront: my agent ran this review; flag anything that reads off and I'll take a personal look.
> 
> This pull request ships the customer subscribe path for authenticated price streams (`prices.crypto`, `prices.crypto.twap`, `prices.equity`, `prices.polymarket`). The production host flip is still deferred. I request changes for two hangs on the default price socket.
> 
> ## Requested changes
> 1. One failed control frame ends every listener on the shared socket.
>    `#flush` calls `#fail` on any `SubscriptionRejectedError` or `TransportError`, including the 10-second ack timeout. `#fail` rejects every key, ends every listener, then closes the connection. A `bad_filter` or a single timeout then kills sibling streams that the server kept. Scope the reject to the batch keys. Treat ack timeout and send failure as a restart.
> 2. `subscribe()` never settles when the first connect fails.
>    `add()` does `void this.#connect().catch(() => this.#schedule(1006))`. Retries continue while keys remain. `state.ready` is never rejected. Callers that await `subscribe()` hang and cannot close a handle they never received. Reject never-subscribed keys after the first failed connect, or after a deadline.
> 
> A related hang: after the last `remove`, a drop in the 1-second idle window leaves `#started === true` and `#authenticated === false`. The next `add()` does not call `#connect()`.
> 
> ## Scope
> The changeset documents that `wss://ws-live-v2.polymarket.com/ws` is not deployed yet. A merge to `main` publishes through Release and canary. Hold the production default, or hold the merge, until that host is live. This is sequencing, not a logic defect in the client.
> 
> `keyTarget` only halves on `dropped` and never recovers. That is a capacity ratchet, not a wrong price.
> 
> ## Notes
> * The branch added `legacy.ts` and then deleted it. That matches the RTDS deletion.
> * No prior review threads.
> * The changeset tells consumers to remove `protocol`, `rtdsLegacy`, and `includeSnapshot`. Those names were not on the prior public surface.
> * The example script requires `POLYMARKET_INTEGRATION_ENVIRONMENT_CONFIG`, so it does not run for an external user on the default production object.

thanks, all were valid findings, fixed in [b56ad85](https://github.com/Polymarket/ts-sdk/pull/351/commits/b56ad85678c757f379e0cbe7d63345c3bbbf685f)

ExoMonk:
@kartojal Review from the PolyBolt side at f99686d2f4c86aac0fa028a570eea4b879486a55, checked against polybolt main e696ec4 (README wire protocol v2, the hosted AsyncAPI at ws-live-v2.polymarket.com/asyncapi.json, RECONNECT.md, src/edge/mod.rs and src/auth/clob.rs), the docs migration guide (Polymarket/docs#388) and the prod deployment state.

**Verdict: request changes.** The client is close to the contract and the transport work since samsondav's review is solid, but one correctness gap on `prices.polymarket`, two migration traps and the release sequencing need to land before this ships.

## Verified against the contract

- Auth op shape `{op:"auth", auth:{apiKey, secret, passphrase}, rid}` matches `ClobCredentials` (protocol.rs); one auth per connection, so the 8-attempt budget is never at risk. `auth_invalid` ends the socket as terminal and `auth_unavailable` restarts with backoff (socket.ts:187-197): both are the non-closing acks the README defines for the CLOB lane.
- Op pacing: one operation frame per 110 ms including auth (socket.ts:216-223) stays under the 20 frames/s window; frames split at 60 KB under the 64 KB cap (socket.ts:248-252); 64 keys per socket with sharding to a second socket (pool.ts:100-110). `{"op":"ping"}` is not metered by the subscribe window (mod.rs:3065), so the 30 s heartbeat is free.
- Reconnect policy (protocol.ts:56-61): 4003 uniform [0, 10 s], everything else full jitter capped at 30 s, 4001 and 4008 terminal with no retry (socket.ts:428). Matches RECONNECT.md.
- Batch acks: the server answers one `subscribed` per item even when items share a channel (mod.rs:2864-2870 loops `subscribe_one` per item, each enqueuing its own ack at mod.rs:3152 or :3162 for idempotent duplicates), so the per-channel countdown in socket.ts:344-351 is correct. Cursor's "duplicate channel acks stall subscribe" finding does not hold; the fixture has no multi-item batch ack to support it.
- Envelope and payload parsing: `full_accuracy_value` preferred and never flattened, equity `received_at` / `is_carried_forward` optional, unknown envelope fields (`id`, `kind`) stripped, unknown channels ignored. The staging fixture confirms the payload symbol spelling is the canonical one (`btcusd`, `aapl`) so the lowercase filter match at socket.ts:368-373 works.
- Credentials never reach logs or error messages; the integration harness redacts auth payloads.

## Requested changes

1. **`prices.polymarket` loses its snapshot.** `parsePolyboltEvent` maps `snapshot: true` to `type: 'subscribe'`, but `PolymarketPriceEventSchema` only accepts `type: 'update'` (polybolt.ts:190-194), so a warm token's snapshot, which per the AsyncAPI carries the current `{market, asset_id, best_bid, best_ask, hash, timestamp}`, fails to parse and is dropped. The unit test at polybolt.test.ts:18-21 codifies dropping every BBO snapshot. Only the cold `[]` barrier should be skipped; a consumer subscribing to an illiquid token otherwise waits for the next trade to see any price. Emit it as a `subscribe` event (or an `update` flagged from the barrier) and add the AsyncAPI "snapshot" example to the fixture-driven tests.

2. **TWAP `windowSeconds: 30` is accepted but can never deliver.** PolyBolt only carries the 60 s TWAP series (AsyncAPI `TwapPoint.window_seconds` const 60; the fixture README itself records the 30 s barrier as empty). price-subscriptions.ts:29 and protocol.ts:5 still offer 30 and the soak subscribes to it. A consumer migrating the old `prices.crypto.chainlink.twap` 30 s stream gets silence with no error. Reject 30 with a `UserInputError` that names 60, and drop the 30 s expectation from the integration and soak tests.

3. **Legacy Binance symbols pass validation and go dark.** The old `prices.crypto.binance` stream took `btcusdt`; PolyBolt symbols are USD pairs with no `t` suffix (docs mapping table), and the server answers an unknown symbol with an empty barrier rather than `bad_filter` (fixture: `btcusdt` snapshot with `data: []`). The changeset only says "feed symbols depend on the deployed source". Either normalize `*usdt` to `*usd` (matching the slash stripping you already do for TWAP) or reject it, and spell the symbol grammar out in the changeset with a link to the catalog in the docs.

4. **Release sequencing.** The production default `wss://ws-live-v2.polymarket.com/ws` is still WAF default-block (infra `Pulumi.polybolt-prod-pmd-euw1.yaml`, allowlist = office plus QA and cluster NAT egress). From outside that list an upgrade is a CloudFront 403, which is what your own validation saw. RTDS is deleted in the same PR, so a release today removes working crypto and equity streams and replaces them with connection failures for every consumer until INFDEFI-1450 clears the gate. Two further server facts matter for the timing: the CLOB credential verifier is not wired on prod yet (`AUTH_CLOB_GRPC_URL` unset; in today's `shadow` mode every trio is admitted, under `enforce` an unwired lane answers `auth_unavailable` to every SDK, which this client turns into a permanent reconnect loop), and the docs migration guide keeps `comments` and `activity` on RTDS because PolyBolt has no channel for them. Recommendation: keep the RTDS transport for `comments` (and as the deprecated path for prices) for one release, ship the PolyBolt lane behind the `realtime` environment entry, and flip the default only after the gate is lifted and the CLOB lane is wired. If product accepts dropping `comments` streaming, say so explicitly in the changeset; today it reads as a side effect.

## Should

5. `prices.polymarket` is a public channel on the server (no auth ever, README "Channel auth classes"), but the SDK requires a secure client for it (subscriptions.ts:373-378). Public-client consumers lose the one price feed they are entitled to. Allow it on the public client with an unauthenticated socket (credentials optional in `PolyboltSocketOptions`, skip the auth request when absent).

6. `dropped` triggers an unsubscribe plus resubscribe of every key on that channel (socket.ts:385-400). `dropped` counts conflated frames skipped for a slow reader; the next frame is already the latest value, so the resubscribe buys nothing for price channels and adds two frames, a fresh two-minute history barrier and a delivery gap to a socket that is already behind. Under sustained slowness that loop feeds the 4002 slow-consumer close. Surface `dropped` on the event (already done) and drop the resync, or limit it to the `keyTarget` halving.

7. A per-item error ack rejects the whole batch (socket.ts:288-292) although the server applies the remaining items ("an unknown-channel item = per-item bad_channel ack, remaining items still apply"); the follow-up unsubscribe cleans up, but sibling consumers in the same batch see a terminal error. Scope the rejection to the item named in `channel` where the code allows it.

## Nits

- The SDK topic names (`prices.crypto`) differ from the wire channel names (`price.crypto`) that docs.polymarket.com and the AsyncAPI use; a mapping line in the JSDoc or README saves a support thread.
- `.changeset` marks the release `minor` while removing streams; fine for 0.x, but the body should list `comments` streaming as removed functionality, not only as a renamed API.
- The protocol integration test sends 65 `price.polymarket` items and expects 4008 (realtime-polybolt.protocol.test.ts:64-76); that closes a real connection on the shared edge every CI run and counts against the per-IP connect budget. Keep it, but behind the same opt-in as the soak.

Prod state for the record (2026-09-11): pods on v0.1.2 with vendor feeds live, `AUTH_TICKET_MODE=shadow`, CLOB lane unwired, public edge gated (`/ws` 403, `/asyncapi.json` 200), `polybolt_sockets` 0.

ExoMonk:
@kartojal Re-review at f72f9b32841bc056b512a14cfc3516c31a3b0d74 (one commit since my review, plus the restored RTDS files, which are byte-identical to `main` except for a `@deprecated` tag).

**Verdict: approve from the PolyBolt side.** Every requested change landed and holds against the contract; what remains is three non-blocking follow-ups.

## Closed

1. BBO snapshot: `PolymarketPriceEventSchema` now accepts `subscribe` (polybolt.ts:190-203), the AsyncAPI warm example is a fixture with its own test, and only the cold `[]` barrier is skipped. Late joiners on a shared key get the latest BBO replayed as a `subscribe` event (socket.ts `refreshSnapshot`, covered by the new socket test).
2. TWAP window: `windowSeconds` is `60` at the type level and at runtime with a migration message (price-subscriptions.ts:29-31); the 30 s cases moved to the invalid list and out of the soak. Matches the AsyncAPI `TwapPoint.window_seconds` const.
3. Symbols: `normalizeCryptoSymbol` lowercases, strips slashes and rewrites a trailing `usdt` to `usd` (protocol.ts:48-50), unit-tested with `BTC/USDT` to `btcusd`; the changeset spells the grammar out and links the catalog page (resolves 200).
4. Sequencing: RTDS is back as the deprecated transport for `comments` and the source-named price topics, PolyBolt topics are opt-in additions, and the changeset says plainly that production access and CLOB verification are still rolling out. A release no longer breaks anyone; consumers who opt into `prices.*` before the gate lifts get a `TransportError` on subscribe, which is the right failure.
5. `prices.polymarket` on the public client rides an unauthenticated socket (`credentials` optional, `#open` skips the auth request), and the type surface narrows the public `realtime` manager to that topic.
6. `dropped` no longer resubscribes; it only records the drop for `keyTarget` placement (socket.ts:430-438), and the integration test asserts no `unsubscribe` after a drop.
7. Per-item errors: an error ack whose `channel` is unambiguous within the pending batch rejects only that item, without queuing an unsubscribe for something the server never registered (socket.ts:364-383, `#flush` rejections path); the accepted sibling resolves. Ambiguous cases fall back to rejecting the batch.

CI is green on the head and Cursor posted no new finding. samsondav's three threads were addressed in b56ad856 and can be resolved.

## Non-blocking follow-ups

- **Stale symbol echo assertion.** `subscriptionsFor` now passes the normalized wire symbol as the echo (protocol.ts:35-45), so a `btc/usd` TWAP subscription reports `payload.symbol === 'btcusd'`. That is the consistent choice (the new unit test asserts it), but the integration test still expects `'btc/usd'` (realtime-polybolt.test.ts:168) and pool.ts:68 still says "including TWAP slashes". The integration suite is not part of PR CI, so this only shows on the next dispatch; align the test and the comment with the new behavior.
- **Positional ack attribution would remove the ambiguous fallback.** The edge applies batch items in order and enqueues exactly one ack per item on the FIFO control lane (mod.rs:2864-2870 per-item loop, acks at :3152/:3162, per-item error acks through the same path), so for a given `rid` the k-th ack always belongs to the k-th item. Matching by position instead of by `channel` makes two same-channel items with one rejection attributable, so no accepted sibling gets torn down and re-unsubscribed. Small change, can be a separate PR.
- `EnvironmentConfig.realtime` is a new required field; it is `@internal`, so fine, just noting it for anyone who builds a full config literal rather than a fork.

Server-side state is unchanged since this morning: prod on v0.1.2, ticket lane `shadow`, CLOB lane unwired, public edge gated. The order that keeps this release safe is: this merges as opt-in, the WAF gate lifts (INFDEFI-1450), the CLOB verifier gets wired (DENG-592), and only then does a later release flip the price topics to PolyBolt by default.

packages/bindings/src/subscriptions/__fixtures__/polybolt/staging.json
```diff
@@ -0,0 +1,4924 @@
+[
```
cesarenaldi: Could we replace this JSON file with representative frames directly in the tests? Our testing approach favors small, focused cases that protect meaningful behavior. Most of these 4,924 lines are repeated history points, which add review overhead without much additional coverage.

There are about nine distinct price-frame cases here. Defining those inline, with one or two history points per snapshot and explicit assertions on the normalized output, would make the contract easier to understand and maintain. A short comment can preserve the capture’s provenance.

packages/client/src/actions/subscriptions.ts
```diff
@@ -212,9 +235,44 @@ export type PublicSubscriptionSpec =
...
+};
+/** Equity updates and recent history. Requires a secure client. */
+export type EquityPriceSubscription = {
+  topic: 'prices.equity';
+  symbol: string;
+  types?: readonly ('subscribe' | 'update')[];
+};
+/** Public best bid and offer snapshots and updates. */
+export type PolymarketPriceSubscription = {
+  topic: 'prices.polymarket';
```
cesarenaldi: Could we clarify the intended user workflow for `prices.polymarket` before exposing it? It appears to overlap with the best bid/ask data already available through the market subscription, so it’s unclear when a consumer should choose one over the other.

I’d expect the direct CLOB stream to have a latency advantage, but we should confirm that with the PolyBolt team rather than assume it. Understanding the latency profile and any differences in delivery guarantees would help us decide whether this deserves a separate public surface and how to explain that choice to users.

packages/client/tests/integration/realtime-fixtures.ts
```diff
@@ -0,0 +1,19 @@
+import { createSecureClient, type SecureClient } from '@polymarket/client';
+import { it as base } from './fixtures';
+
+// Price streams only need signer authentication. Using the EOA avoids wallet
+// deployment and relayer authorization in these non-trading integration tests.
+export const it = base.extend<{ realtimeClient: SecureClient }>({
```
cesarenaldi: Could we reuse `secureClientWithDepositWallet` here and remove the additional fixture? Realtime authentication uses the client’s API credentials, and the subscription integration test already uses that fixture for `prices.crypto`. Is there a specific setup constraint that prevents these tests from doing the same?

packages/client/tests/integration/realtime-polybolt.test.ts
```diff
@@ -0,0 +1,353 @@
...
+  }) => {
+    vi.stubGlobal('WebSocket', ObservedWebSocket);
+    try {
+      const handle = await client.subscribe([
+        { topic: 'prices.crypto', symbols: ['btcusd', 'ethusd'] },
+      ]);
+      const socket = ObservedWebSocket.connections[0];
+      expect(socket).toBeDefined();
+      // Inject only the loss signal; subsequent recovery uses the real server.
+      socket?.dispatchEvent(
```
cesarenaldi: Could we move the synthetic dropped-frame and close-event cases into the controlled subscription tests? Passive observation of real WebSocket traffic makes sense here for checking connection sharing, but injecting server events mixes simulated failures with live integration coverage. Those cases could run deterministically without depending on API access, while this suite verifies the SDK against actual server behavior.

packages/client/src/actions/subscriptions.ts
```diff
@@ -212,9 +235,44 @@ export type PublicSubscriptionSpec =
...
+ */
+export type CryptoPriceSubscription = {
+  topic: 'prices.crypto';
+  symbols: readonly string[];
+};
+/** 60-second time-weighted prices and recent history. Requires a secure client. */
+export type CryptoTwapPriceSubscription = {
+  topic: 'prices.crypto.twap';
+  symbols: readonly string[];
+  windowSeconds: 60;
```
cesarenaldi: Could we remove `windowSeconds` from the subscription input? With only `60` supported, this is a required option with no choice: it adds a parameter without letting the caller express any additional intent. Let’s set the 60-second window internally and document it on the topic. We can retain `windowSeconds` in the returned events as descriptive metadata.

packages/client/src/actions/subscriptions.ts
```diff
@@ -119,26 +140,28 @@ export type SportsSubscription = {
...
 
+/** @deprecated Compatibility stream. Prefer HTTP comment APIs for new integrations. */
 export type CommentsSubscription = {
   topic: 'comments';
   types?: readonly CommentsEventType[];
   parentEntityId?: number;
   parentEntityType?: 'Event' | 'Market';
 };
 
+/** @deprecated Use {@link CryptoPriceSubscription}. */
```
cesarenaldi: Could we expand this deprecation TSDoc with the replacement topic and an inline migration example? Consumers need to see that the replacement requires an authenticated client, explicit symbols, and handling history snapshots. For example:

````ts
/**
 * @deprecated Use `prices.crypto` with a secure client and explicit symbols.
 * Symbols are normalized to USD pairs, and events include history snapshots.
 *
 * @example
 * ```ts
 * // Before
 * const legacyPrices = await publicClient.subscribe([
 *   { topic: 'prices.crypto.chainlink', symbols: ['btc/usd'] },
 * ]);
 *
 * // After — secureClient is already authenticated
 * const prices = await secureClient.subscribe([
 *   { topic: 'prices.crypto', symbols: ['btcusd'] },
 * ]);
 *
 * for await (const event of prices) {
 *   if (event.type !== 'update') continue;
 *   console.log(event.payload.symbol, event.payload.value);
 * }
 * ```
 */
````

Please also cover the deprecated Binance topic and explicitly distinguish USDT from USD rather than presenting that as a like-for-like replacement. Let’s state the planned removal release and include this guidance in the migration docs, since inline topic literals don’t surface the containing type’s deprecation warning.

File: packages/client/src/actions/subscriptions.ts [hunk 10]
```diff
@@ -119,26 +140,28 @@ export type SportsSubscription = {
...
 export type CryptoPricesSubscription = {
   topic: CryptoPricesTopic;
   symbols?: readonly string[];
 };
 
+/** @deprecated Use {@link CryptoTwapPriceSubscription}. */
```
cesarenaldi: Could we add a before/after migration example directly to this type’s deprecation TSDoc? The supported migration is the 60-second workflow; the existing 30-second option has no corresponding replacement. Following the requested removal of the fixed `windowSeconds` input from the new topic, the example could be:

````ts
/**
 * @deprecated Use `prices.crypto.twap` with a secure client and explicit symbols.
 * The replacement provides 60-second TWAP prices and history snapshots.
 * There is no replacement for the 30-second window.
 *
 * @example
 * ```ts
 * // Before
 * const legacyPrices = await publicClient.subscribe([
 *   {
 *     topic: 'prices.crypto.chainlink.twap',
 *     symbols: ['btc/usd'],
 *     windowSeconds: 60,
 *   },
 * ]);
 *
 * // After — secureClient is already authenticated
 * const prices = await secureClient.subscribe([
 *   { topic: 'prices.crypto.twap', symbols: ['btcusd'] },
 * ]);
 *
 * for await (const event of prices) {
 *   if (event.type !== 'update') continue;
 *   console.log(event.payload.symbol, event.payload.value);
 * }
 * ```
 */
````

Please keep the example aligned with the final input contract, state the planned removal release, and include this migration in the shared migration docs.

File: packages/client/src/actions/subscriptions.ts [hunk 11]
```diff
@@ -119,26 +140,28 @@ export type SportsSubscription = {
...
 export type CryptoPricesChainlinkTwapSubscription = {
   topic: CryptoPricesChainlinkTwapTopic;
-  /** Averaging window used to calculate each TWAP price. */
   windowSeconds: CryptoPricesChainlinkTwapWindowSeconds;
-  /** Lowercase slash-delimited symbols, such as `btc/usd`. */
   symbols?: readonly string[];
 };
 
+/** @deprecated Use {@link EquityPriceSubscription}. */
```
cesarenaldi: Could we add an inline migration example to this type’s deprecation TSDoc, showing the new topic and authenticated client while preserving the caller’s event filter? For example:

````ts
/**
 * @deprecated Use `prices.equity` with a secure client.
 * The existing symbol and event-type filters remain available.
 *
 * @example
 * ```ts
 * // Before
 * const legacyPrices = await publicClient.subscribe([
 *   { topic: 'prices.equity.pyth', symbol: 'aapl', types: ['update'] },
 * ]);
 *
 * // After — secureClient is already authenticated
 * const prices = await secureClient.subscribe([
 *   { topic: 'prices.equity', symbol: 'aapl', types: ['update'] },
 * ]);
 * ```
 */
````

Please also state the planned removal release and include this example in the shared migration docs. Keeping the full example at the original declaration is enough; it doesn’t need duplication on decorator re-exports.

File: packages/client/src/websockets/realtime/protocol.ts [hunk 12]
```diff
@@ -0,0 +1,65 @@
...
+            window_seconds: spec.windowSeconds,
+          }
+        : { symbol },
+      symbol,
+    );
+  });
+}
+
+function normalizeCryptoSymbol(input: string): string {
+  return input.toLowerCase().replaceAll('/', '').replace(/usdt$/, 'usd');
```
cesarenaldi: Could we require the canonical symbol format for the new topics instead of accepting and rewriting legacy spellings? We’re already providing migration guidance, so callers can explicitly switch to `btcusd`. Please reject inputs such as `btc/usd` and `btcusdt` with a clear validation error rather than silently converting them. In particular, USDT → USD changes the quote currency, not just the formatting. The deprecated topics can retain their existing behavior.

packages/client/src/clients.ts
```diff
@@ -176,7 +177,7 @@ abstract class AbstractClient<TContext extends PublicContext> {
     return this.context.perps;
   }
 
-  /** @internal */
+  /** Shared managers for realtime subscriptions. */
```
cesarenaldi: Please keep `webSockets` marked `@internal` here and on the secure-client override. The transport managers are implementation details; consumers should use `client.subscribe()` through the unified SDK surface. Please also remove the new root exports of `RealtimeWebSocketManager`, `RtdsWebSocketManager`, and their options from the subscriptions decorator so we don’t introduce a second public subscription interface.

packages/client/src/actions/subscriptions.test-d.ts
```diff
@@ -1,46 +1,161 @@
...
+    >().toEqualTypeOf<EquityPriceEvent>();
+    expectTypeOf<
+      EventForSubscriptionSpecs<
+        [{ topic: 'prices.polymarket'; assetIds: ['1'] }]
+      >
+    >().toEqualTypeOf<PolymarketPriceEvent>();
+    const pending = secureClient.subscribe([
+      { topic: 'prices.crypto', symbols: ['btcusd'] },
+    ]);
+    expectTypeOf(pending).resolves.toEqualTypeOf<
```
cesarenaldi: Could we extend the direct `client.subscribe()` type tests to cover TWAP, equity, and BBO, as we already do for crypto? Each single-topic subscription should infer only its corresponding event type, including its own snapshot/update variants. Please also add a mixed-topic case to verify that the result includes only the requested topics. These checks should exercise the public client method, not just the helper type mappings.

packages/client/src/errors.ts
```diff
@@ -327,3 +331,15 @@ export function makeErrorGuard<
     },
   };
 }
+
+/** A realtime subscription was rejected by the server. */
+export class SubscriptionRejectedError extends PolymarketError {
```
cesarenaldi: Could we reuse `RequestRejectedError` for server-rejected subscriptions instead of introducing another public error type? We already use it for WebSocket rejections in Perps, and it carries a machine-readable code.

This would also remove the discrepancy introduced by `error.channel`: callers subscribe to `prices.crypto`, but the proposed error reports the wire channel `price.crypto`. We don’t need to expose that additional transport detail; `RequestRejectedError` and its code should give consumers enough information to act on the rejection. If subscriptions need a distinct error, could we clarify what additional consumer behavior requires it?

packages/client/src/actions/subscriptions.ts
```diff
@@ -119,26 +140,28 @@ export type SportsSubscription = {
   topic: 'sports';
 };
 
+/** @deprecated Compatibility stream. Prefer HTTP comment APIs for new integrations. */
```
cesarenaldi: If this stream is intended to be discontinued, please state that explicitly in the deprecation TSDoc and direct consumers to `client.listComments()` with an example:

````ts
/**
 * @deprecated The comments stream is being discontinued.
 * Use `client.listComments()` to retrieve comments.
 * This method does not provide live comment or reaction events.
 *
 * @example
 * ```ts
 * const page = await client.listComments({
 *   parentEntityId: '123',
 *   parentEntityType: CommentParentEntityType.Event,
 *   pageSize: 20,
 * }).firstPage();
 *
 * for (const comment of page.items) {
 *   console.log(comment);
 * }
 * ```
 */
````

Please also include the planned removal release once confirmed.

File: packages/client/tests/integration/realtime-polybolt.soak.test.ts [hunk 17]
```diff
@@ -0,0 +1,93 @@
+import { describe, expect, vi } from 'vitest';
+import { it } from './realtime-fixtures';
+
+describe.skipIf(process.env.POLYMARKET_REALTIME_SOAK !== '1')(
+  'realtime price soak',
+  () => {
+    it('streams 100 filters for 30 minutes', async ({
```
cesarenaldi: This appears to be an end-to-end soak test of the SDK, network, and deployed realtime service rather than a focused SDK regression test. Please coordinate with the QA team to move it into the appropriate end-to-end test suite and remove it from the SDK tests, along with its dedicated CI option. We should keep focused integration coverage for SDK subscription behavior here.

cesarenaldi: When removing the soak test and its CI option, please also address the `POLYMARKET_REALTIME_SOAK` gate in `realtime-polybolt.protocol.test.ts`. If these protocol checks remain SDK integration coverage, they should run with the normal integration suite rather than depend on that flag. Otherwise, please move them with the corresponding QA coverage so we don’t retain a permanently skipped suite.

packages/client/src/actions/subscriptions.ts
kartojal: prices.polymarket will not be supported

packages/client/src/actions/subscriptions.ts
kartojal: added, i will add the planned removal date once is released and we are certain of the date

packages/client/src/errors.ts
kartojal: done, reused `RequestRejectedError`

cesarenaldi (approved):
The requested SDK changes are addressed. Build, lint, typecheck, and 589 unit/type tests pass.

Production streaming remains unverified from my environment: client setup succeeds, but the WebSocket upgrade returns HTTP 403 before authentication or event delivery.

kartojal:
> The requested SDK changes are addressed. Build, lint, typecheck, and 589 unit/type tests pass.
> 
> Production streaming remains unverified from my environment: client setup succeeds, but the WebSocket upgrade returns HTTP 403 before authentication or event delivery.

Production is verified via internal nets, but currently is gated until release.

ExoMonk:
**Round 3, head e60506ef6457c579e4f4b1f19ffd233b76a0ddcb. Verdict: APPROVE for release, with one release-order item that is on our side (docs) and two non-blocking nits.**

Re-read the 11 commits since my approval at f72f9b32 against the PolyBolt contract as it runs on prod today: `wss://ws-live-v2.polymarket.com/ws`, `AUTH_TICKET_MODE=enforce`, CLOB Auth wired over TLS, edge public since 2026-09-15 ~18:30Z.

### Contract check of the transport split

- `PolyboltConnection` emits exactly the frames the server parses: `{op:"auth", auth:{apiKey,secret,passphrase}, rid}` and `{op:"subscribe"|"unsubscribe", subscriptions:[{channel, filter}], rid}` with `price.crypto {symbol}`, `price.crypto.twap {symbol, window_seconds:60}`, `price.equity {symbol}`. Checked against `src/protocol.rs` and the edge parser on polybolt main.
- Pacing moved from the session flush to the connection: capacity one at 110 ms, covering auth and every op frame, so at most ~9 ops/s against the server's 20 ops/s budget (ping is unmetered). The ack timer now starts when the frame is actually sent rather than when it is queued, which is the correct side of the pacing delay.
- Per-item error attribution is unchanged: an ambiguous rejection (two items on the same channel) still rejects the whole batch, and the session enqueues an unsubscribe for the siblings the server kept, so client and server converge.
- Close codes: 4001 and 4008 are terminal, everything else including the 4003 drain reconnects with the draining jitter policy. Right for a surface that never sends tickets (4001 is the ticket lane's hostile-shape close).
- `RealtimeErrorCode` matches the server's ten codes in `src/protocol.rs` one for one.
- The `seq` wording in the changeset and TSDoc is accurate: scoped per channel per connection, resets on reconnect.
- Symbol policy `^[a-z0-9]+usd$` is consistent with what prod serves: `price.crypto` symbols are the Pyth `Crypto.X/USD` feeds lowered to `xusd` (`btcusd` in the AsyncAPI example and in the staging canary). Note that the SDK now hard-codes a quote-currency rule the server does not have; a non-USD-quoted pair added to the feeder would be refused client-side. Acceptable while the catalog is operator-controlled.

### Findings

1. **Release order, docs side (mine to fix).** The changeset links to `https://docs.polymarket.com/api-reference/live-data/migrating-from-rtds#migrate-sdks-to-polybolt`. That page is docs #388, still unmerged (the URL is a 404 today), and the page has no "Migrate SDKs to PolyBolt" section: its headings are What changes, Topic mapping, Before and after, Client checklist, What stays on RTDS. Added in docs #388 at eaa2b52f6cebb20433fffb818ea2342d140a56f0 (EN and CN), slug `migrate-sdks-to-polybolt`, matching the final API here: secure client required, canonical `btcusd` symbols, fixed 60 s TWAP window, `RequestRejectedError.code`, per-connection `seq`, removal one month after 0.11.0, no SDK surface for `price.polymarket`. Docs #388 has to merge before the SDK release or the changelog link is dead on arrival.

2. Nit, forward compatibility. `PolyboltAckSchema.code` is `z.enum(RealtimeErrorCode)`. A future server code the SDK does not know makes the whole ack unparseable, so the pending op times out after 10 s and the session restarts instead of surfacing a rejection. `z.string()` on the wire schema (keeping the enum as the known set) would degrade to a clean `RequestRejectedError` carrying the unknown code. Not blocking: the ten codes are stable.

3. Nit. `PriceSession.#event` still carries the `'assetId' in subscription` branch from `prices.polymarket`. `PriceKey` has no `assetId` member any more, so it is dead code that only compiles because `in` narrows to `unknown`.

### Threads and Cursor

- Cursor's single finding (bcdb21b, poll timeouts in the protocol test) is fixed at head: all four `expect.poll` calls carry `timeout: 10_000`. No Cursor findings on e60506ef64.
- The three samsondav threads on socket.ts are addressed by the session split (batch-scoped rejection with restart on ack timeout; an open failure rejects never-accepted keys; `add()` reconnects whenever the session is not authenticated, connecting or scheduled), each with a session.test.ts case. Cesar's TWAP deprecation TSDoc thread is addressed (before/after example present). All four can be resolved.

### On `prices.polymarket` (Cesar's latency question)

Measured today from the public internet, so the "confirm rather than assume" ask on latency has an answer. Two sockets from one machine, subscribed to the same 16 outcome tokens (the 8 most-traded markets): PolyBolt `price.polymarket` and the CLOB market socket `wss://ws-subscriptions-clob.polymarket.com/ws/market`. Events are matched on `(asset_id, hash)`: the CLOB book hash travels verbatim through `data.v1.PriceChange` into PolyBolt, so an equal hash is the same event. 150 s window starting 2026-09-15T19:07:06Z, 4,546 PolyBolt updates.

| | PolyBolt `price.polymarket` | CLOB market socket |
| --- | --- | --- |
| Arrival relative to the other socket, same event | p50 +15 ms, p90 +48 ms, p99 +129 ms (min -474, max +501, mean -3) | reference |
| CLOB event time to client (e2e) | p50 55 ms, p90 86 ms, p99 216 ms | p50 38 ms, p90 142 ms, p99 450 ms |
| Best bid/ask on matched events | identical on 4,546 of 4,546 | |
| PolyBolt events with no CLOB counterpart | 0 | |

A 90 s rerun at 19:11:56Z gave the same picture: 5,034 updates, arrival p50 +14 ms, p90 +39 ms, p99 +75 ms; e2e p50 60 / p90 83 / p99 128 ms against p50 45 / p90 78 / p99 163 ms on the CLOB socket; values identical on 5,034 of 5,034; zero PolyBolt events without a CLOB counterpart.

So the public channel sits about 15 ms behind the direct CLOB socket at the median, with a tighter tail, and never disagrees on the values. The counterpart asymmetry is structural: the CLOB socket lists every level change inside one book update with its own intermediate hash (4,626 such hashes in the rerun) while PolyBolt carries one record per book update, the final state. Two consequences for anyone consuming it: PolyBolt emits on every book update, not only when the best bid or ask moves (about 56 frames/s across these 16 tokens), so a consumer dedupes unchanged values itself; and a PolyBolt subscriber sees the same best bid/ask as a `market` subscriber, roughly 15 ms later. That supports keeping `prices.polymarket` out of the SDK for now: next to the existing `market` subscription it adds no information and no latency advantage worth a second public surface. Probe script and raw output available on request.

### Before tagging

- Run the metered suite once against prod from outside the VPN (`POLYMARKET_RUN_METERED_TESTS=1` with the builder trio). The edge is public and enforce is live; the CLOB lane is proven from the public internet (a deliberately invalid trio answers `auth_invalid` in about 150 ms).
- Merge docs #388 first (finding 1).

